PR #57, opened 2014-03-12
No need to be root for pcapplay.

wdoekes:
The first part, needed to get pcapplay without root powers going.

### DIFF
--- a/include/send_packets.h
+++ b/include/send_packets.h
@@ -114,6 +114,7 @@ typedef struct {
     /* Used in send_packets thread */
     struct sockaddr_storage to;
     struct sockaddr_storage from;
+    int sock;
 } play_args_t;
 
 #ifdef __cplusplus
--- a/src/call.cpp
+++ b/src/call.cpp
@@ -244,11 +244,12 @@ uint16_t get_remote_port_media(const char *msg, int pattype)
  */
 void call::get_remote_media_addr(char *msg)
 {
-    uint16_t video_port, audio_port;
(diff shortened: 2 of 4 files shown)


vodik:
Another option is to set CAP_NET_RAW on sipp.

wdoekes:
An interesting option, but it has its own drawbacks.

I tried it indeed:

    sudo setcap cap_net_raw+ep `which sipp`

But that broke my unittest framework, which introspects which sockets are open by looking at proc:

    $ ls /proc/`pidof sipp`/fd
    ls: cannot open directory /proc/9986/fd: Permission denied

vodik:
Your personal unittest stuff Linux? Maybe `ss` would be more appropriate. The `-p` flag adds process information to the output.

wdoekes:
Same difference:

    $ ss -lp | grep 5060
    tcp    UNCONN     0      0                    *:ipproto-5060               *:*        users:(("sipp",3600,4))

Vs:

    $ ss -lp | grep 5060
    tcp    UNCONN     0      0                    *:ipproto-5060               *:*

Was to be expected ;)

vodik:
Yeah, huh. TIL.

wdoekes:
Interestingly, the polycom-sipped also did some work on this, seeing that the RAW sockets changed to DGRAM in this commit:
https://github.com/SIPp/polycom-sipped/commit/e9b1f62ca1efbf0be010b2ea0650c601a43f1198

alienpenguin:
hi, is there any update on this PR?

PR #74, opened 2014-05-28
send_packets: Avoid crashing on errors in sendto call.

sruffell:
If sendto() returned an error, pthread_cleanup_pop() was not called before
return, resulting in a segfault on my CentOS 5.x platform.

Next, I need to look into why sendto is failing on CentOS 5.x but not on other
platforms.

### DIFF
--- a/src/send_packets.c
+++ b/src/send_packets.c
@@ -238,9 +238,8 @@ int send_packets (play_args_t * play_args)
         }
 #endif
         if (ret < 0) {
-            close(sock);
             WARNING("send_packets.c: sendto failed with error: %s", strerror(errno));
-            return( -1);
+            break;
         }
 
         rtp_pckts_pcap++;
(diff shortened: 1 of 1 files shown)


tebriel:
I was seeing the same segfault issue, randomly after 30-70 calls. I'm running in two dockers communicating to each other over a bridged network. 

dmesg shows:
```
[2860961.852761] traps: sipp[9672] general protection ip:7f6e211a32f0 sp:1f45fa96183e4bc9 error:0 in libc-2.23.so[7f6e2116e000+1bf000]
```

And the error I see in sipp is:
```
2016-11-09      20:36:04.498133 1478723764.498133: send_packets.c: sendto failed with error: Operation not permitted.
```

Sipp was exiting status code 139.

I can reproduce with great regularity, let me know if I can provide any more debugging output.

I created a patch for 2.5.1 since this branch has merge conflicts, patch attached.

[send_packets.patch.txt](https://github.com/SIPp/sipp/files/581861/send_packets.patch.txt)

wdoekes:
Fixed in recent commits

PR #82, opened 2014-07-24
Added play_dtmf logic taken from http://sourceforge.net/p/sipp/patches/50/

horacimacias:
I took the work done by Dmitry Kunilov and dev-jam from http://sourceforge.net/p/sipp/patches/50/ and updated it to the current sipp version. So far things seem to work for me.


vodik:
I have some use for this, actually. We're currently doing dtmf testing with pcap_play.

moises-silva:
yep, this is much better than requiring a pcap to play dtmf

wdoekes:
This has been fixed by #141.

Issue #29, opened 2013-10-07
Better log when receiving unexpected response but expecting request


rkday:
SIPped, the Polycom fork of SIPp (https://github.com/SIPp/polycom-sipped) improves the error message when an unexpected response arrives but a request is expected. It would be good to get this merged into the main SIPp code.

It looks as though the relevant commit is 525494a.

PR #92, opened 2014-09-23
Implemented better log when receiving unexpected response #29

snapschott:


wdoekes:
Closing this out. It might be a useful feature, but we can't tell unless we see it in action.

Feel free to reopen if you clean up the patch.

PR #96, opened 2014-09-29
Added -noscreen option to disable status and result screen.

kovacsbalu:


vodik:
Yeah, if we want an explicit flag, this should be redone ontop of #130 because this also disables signal handling.

wdoekes:
Thanks for the initial pull request though. It's appreciated!

I think we can do without a command line option though, right?

wdoekes:
(fed347b was superseded by c2381c8 -- fixed log message)

kovacsbalu:
Yep, I'll check it! :)

PR #97, opened 2014-09-29
Disable SDP detection when you use skipsdp="true" attribute on reInvite ...

kovacsbalu:
...receive tag.

### DIFF
--- a/include/scenario.hpp
+++ b/include/scenario.hpp
@@ -105,6 +105,7 @@ public:
     int		 counter;
     double         lost;
     int            crlf;
+    int            skipsdp;
     bool           hide;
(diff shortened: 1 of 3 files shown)


wdoekes:
I'm sorry for not replying to this sooner, but why would we want this?

- I don't see an immediate benefit.
- There is no documentation provided for the new feature.
- Code style is violated.

kovacsbalu:
SIPp use sdp from reInvite but when you reject this reInvite you need to use the original sdp.
With this flag you can skip sdp from the reInvite.

wdoekes:
Thanks for the pull request! Closing because #142 supersedes it.

PR #107, opened 2014-10-06
Separate call rate increase to a dedicated task

bklang:
For several load tests, I want to have finer-grained statistics coming from SIPp, so I have the `-fd` option set to `1s`.  However, this means that if I want to use the automatic rate increases, I'm stuck with either coarser stats, or a less-useful ramp-up time.

This PR is an attempt to move call rate increases into a separate task.

Note: I have not yet been able to test this because I'm having trouble with the autotools. I'll put those details into a comment.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.o
 *~
+.*.sw*
 .deps/
 Makefile
 autom4te.cache/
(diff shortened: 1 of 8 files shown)


benlangfeld:
Continued at https://github.com/SIPp/sipp/pull/126.

PR #118, opened 2015-01-20
Fix out-of-tree build

Lekensteyn:
The include directories must refer to the source dir, not the objdir.


rkd-msw:
Looks pretty sensible - thanks!

rkday:
Looks sensible - thanks!

PR #120, opened 2015-01-20
Apply consistent formatting in configure.ac and Makefile.am

Lekensteyn:
Apply consistent use of four-spaces in configure.ac and reflow the
help2man line.

Apply consistent use of tabs in Makefile.am and convert line such as:

    foo_SOURCES = foo.c \
        bar.c

to:

    foo_SOURCES = \
        foo.c \
        bar.c
---
This is a very trivial change and looks evasive, but actually has no functional changes. Consistency makes things easier to read, hence this patch. Later on such changes are not necessary.

There are other source files which have tabs and spaces intermixed, but there are also other style inconsistencies (spacing around `*`: `char *       foo` vs. `char      * foo` vs `char     *foo`). Those either need a one-time fixup or needs to be left alone I guess.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,28 +8,41 @@ AM_CFLAGS=-I$(srcdir)/include
 AM_CPPFLAGS=$(AM_CFLAGS)
 
 CLEANFILES =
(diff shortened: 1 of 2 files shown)


Lekensteyn:
Superseded by #121

PR #121, opened 2015-01-21
Fix `make dist` target

Lekensteyn:
One patch applies consistent formatting, the other one fixes the `make dist` target.

Now you can run `make distcheck` to produce a tarball suitable for distribution.


Lekensteyn:
Rebased on master.

@rkday What do you think of this patch?

rkday:
Looks good - sorry for the delay in merging!

Issue #87, opened 2014-09-19
wrong response computation in auth header


emabo:
there is a problem in createAuthResponseMD5 (file src/auth.cpp). It seems that changeset ff340b80 added an issue when computing a response. Also the comment of the changeset seems to suggest something wrong:
"verifyAuthHeader now takes auth-int into account, but the response is now calculated wrongly"

So, I propose following change to modify the behaviour:

diff --git a/src/auth.cpp b/src/auth.cpp
index ecd1b9e..5add2c1 100644
--- a/src/auth.cpp
+++ b/src/auth.cpp
@@ -260,7 +260,6 @@ int createAuthResponseMD5(const char *user,
     unsigned char ha1_hex[HASH_HEX_SIZE+1], ha2_hex[HASH_HEX_SIZE+1];
     char tmp[MAX_HEADER_LEN];
     md5_state_t Md5Ctx;
-    const char authint[] = "auth-int";

     // Load in A1
     md5_init(&Md5Ctx);
@@ -307,7 +306,7 @@ int createAuthResponseMD5(const char *user,
         md5_append(&Md5Ctx, (md5_byte_t *) ":", 1);
         md5_append(&Md5Ctx, (md5_byte_t *) cnonce, strlen(cnonce));
         md5_append(&Md5Ctx, (md5_byte_t *) ":", 1);
-        md5_append(&Md5Ctx, (md5_byte_t *) authint, strlen(authint));
+        md5_append(&Md5Ctx, (md5_byte_t *) authtype, strlen(authtype));
     }
     md5_append(&Md5Ctx, (md5_byte_t *) ":", 1);
     md5_append(&Md5Ctx, (md5_byte_t *) &ha2_hex, HASH_HEX_SIZE);

rkday:
Sorry - I do agree that this was a valid bug (which you'd only hit if you had `cnonce` but not `qop="auth-int"`, which is why I missed it). I've just accepted #123 which should fix it.

PR #123, opened 2015-02-19
Fix invalid authentication response calculation

Lekensteyn:
The digest calculation when "qop" is given follows from:

    HA1=MD5(username:realm:password)
    HA2=MD5(method:digestURI)
    response=MD5(HA1:nonce:nonceCount:clientNonce:qop:HA2)

When qop is not given, the response is:

    response=MD5(HA1:nonce:HA2)

Currently the qop parameter is hard-coded to "auth-int" which fails to
compute the correct hash for qop=auth. Fix this by using the correct
authentication type instead.

References:

 - https://tools.ietf.org/html/rfc3261#section-22.4
 - https://en.wikipedia.org/wiki/Digest_access_authentication#Overview

Fixes https://github.com/SIPp/sipp/issues/87

### DIFF
--- a/src/auth.cpp
+++ b/src/auth.cpp
@@ -264,7 +264,6 @@ int createAuthResponseMD5(const char *user,
     unsigned char ha1_hex[HASH_HEX_SIZE+1], ha2_hex[HASH_HEX_SIZE+1];
     char tmp[MAX_HEADER_LEN];
     md5_state_t Md5Ctx;
-    const char authint[] = "auth-int";
 
     // Load in A1
     md5_init(&Md5Ctx);
@@ -311,7 +310,7 @@ int createAuthResponseMD5(const char *user,
         md5_append(&Md5Ctx, (md5_byte_t *) ":", 1);
         md5_append(&Md5Ctx, (md5_byte_t *) cnonce, strlen(cnonce));
         md5_append(&Md5Ctx, (md5_byte_t *) ":", 1);
-        md5_append(&Md5Ctx, (md5_byte_t *) authint, strlen(authint));
+        md5_append(&Md5Ctx, (md5_byte_t *) authtype, strlen(authtype));
     }
     md5_append(&Md5Ctx, (md5_byte_t *) ":", 1);
     md5_append(&Md5Ctx, (md5_byte_t *) &ha2_hex, HASH_HEX_SIZE);


Lekensteyn:
Shell script that was used to calculate it by hand from the header:
```sh
#!/bin/sh
set -e -u

# Example Proxy-Authorization:
#Digest username="*1337",realm="cy.sip.incenova.com",cnonce="6b8b4567",
#nc=00000001,qop="auth",uri="sip:*9000@cy.sip.incenova.com",
#nonce="c7903793-a86b-4129-98f0-e8893d44f3c1",
#response="0ea0ce64854f6d6251a051a0572d75aa",algorithm=MD5
if [ $# -eq 0 ]; then
    echo "Usage: echo Proxy-Authorization: ... | $0 password"
    exit 1
fi

username=
realm=
password=$1
method=INVITE
digestURI=
nonce=
nonceCount=
clientNonce=
qop=

DEBUG() {
    : "$@" >&2
}

while read k v; do
    v=$(tr -d \" <<<"$v")
    DEBUG echo "# $k=$v"
    case $k in
    username|realm|qop|nonce) eval "$k=\$v" ;;
    cnonce) clientNonce=$v ;;
    nc)     nonceCount=$v ;;
    uri)    digestURI=$v ;;
    *)  DEBUG echo "ignored $k" ;;
    esac
done < <(tr ', =' '\n\n ')

# HA1=MD5(username:realm:password)
# HA2=MD5(method:digestURI)
# response=MD5(HA1:nonce:HA2)
# response=MD5(HA1:nonce:nonceCount:clientNonce:qop:HA2)    # when qop is given

MD5() {
    for arg; do DEBUG printf ' "%s"' "$arg"; done
    DEBUG echo
    printf "$@" | md5sum | cut -d' ' -f1 | tr -d '\n'
}

HA1=$(MD5 '%s:%s:%s' "$username" "$realm" "$password")
HA2=$(MD5 '%s:%s' "$method" "$digestURI")

if [ -z "$qop" ]; then
    MD5 "%s:%s:%s" "$HA1" "$nonce" "$HA2"
else
    MD5 "%s:%s:%s:%s:%s:%s" \
        "$HA1" "$nonce" "$nonceCount" "$clientNonce" "$qop" "$HA2"
fi
echo
```

rkday:
Thanks! I think there's one edge case where this doesn't work (if qop is "auth,auth-int", which is valid on the 401 but not on the subsequent REGISTER) - but I'd need to double-check that, and I think this patch does improve on the current behaviour.

PR #125, opened 2015-03-09
Replace xp_replace with more correct html escape handling

vodik:
One of our internal tools started spits out sipp scripts containing the `&quot;` html escape which sipp is not capable of handling. This patch set replaces the `xp_replace` with temporary buffer with a new mechanism which is one step closer to a proper html escape parser. The new code path only handles a few more sequences (`&amp;`, `&gt;`, `&lt;`, and `&quot;`) but is a good starting place for a better mechanism.

Another option could be to pull out the xp_parser.c file and replace it with a better, standard, off-the-shelf xml parser.

I've extensively tested this personally, but not sure how to trigger the `xp_parser_ut.cpp` code.

### DIFF
--- a/include/xp_parser.h
+++ b/include/xp_parser.h
@@ -23,7 +23,7 @@
 extern "C" {
 #endif
 
-int   xp_replace(const char *source, char *dest, const char *search, const char *replace);
+int   xp_unescape(const char *source, char *dest);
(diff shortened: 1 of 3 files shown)


wdoekes:
> but not sure how to trigger the xp_parser_ut.cpp code.

That was not so hard ;)

Thanks for the patch!

PR #128, opened 2015-04-26
Add command-line option to allow triple slashes in Call-IDs

rkday:


src/sipp.cpp
```diff
@@ -234,6 +234,7 @@ struct sipp_option options_table[] = {
      "- On any other unexpected message, abort the call by sending a BYE or a CANCEL\n",
      SIPP_OPTION_UNSETFLAG, &default_behaviors, 1},
     {"pause_msg_ign", "Ignore the messages received during a pause defined in the scenario ", SIPP_OPTION_SETFLAG, &pause_msg_ign, 1},
+    {"callid_slash", "Allow triple-slash in SIP Call-IDs.", SIPP_OPTION_SETFLAG, &callidSlash, 1},
```
wdoekes: This comment really only makes sense if you know what the triple slash does in the first place.

Perhaps it should've been:
"Don't give special meaning to triple-slash in SIP Call-IDs"

Issue #42, opened 2013-12-17
Add fax/t38 support 


ctradu:
The guys at https://code.osso.nl/projects/sipp have a nice fork of sipp 3.2 that adds support for t38 fax in sipp.

A new directive send_pcap_image is added and the support for the pcap files containing fax images.

With the permission of the authors, we could add this feature in sipp.

PR #129, opened 2015-05-05
Add fax/t38 support

wdoekes:
See #42 .

This is implemented without the no-root pcap_play requested in #57; that will have to wait for another day.

It was tested by creating fax test setup with the sipp-osso client and then moving to this branch.

### DIFF
--- a/src/rtpstream.cpp
+++ b/src/rtpstream.cpp
@@ -886,7 +886,8 @@ int rtpstream_get_videoport (rtpstream_callinfo_t *callinfo)
 }
 
 /* code checked */
-void rtpstream_set_remote (rtpstream_callinfo_t *callinfo, int ip_ver, char *ip_addr, int audio_port, int video_port)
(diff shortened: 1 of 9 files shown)


src/rtpstream.cpp
```diff
@@ -886,8 +886,10 @@ int rtpstream_get_videoport (rtpstream_callinfo_t *callinfo)
 }
 
 /* code checked */
-void rtpstream_set_remote (rtpstream_callinfo_t *callinfo, int ip_ver, char *ip_addr, int audio_port, int video_port)
+void rtpstream_set_remote (rtpstream_callinfo_t *callinfo, int ip_ver, char *ip_addr,
+                           int audio_port, int image_port, int video_port)
 {
+  // TODO: we don't do anything with image_port right now..
```
rkday: Worth a note to say what we should do with it in future (and what the impact of not doing anything is)?

rkday:
One comment, but otherwise looks good - merge at will.

Issue #75, opened 2014-05-29
Startup failure when running from environment without TERM.


wdoekes:
If you run sipp in foreground mode, it will `screen_init`.

This fails with the following error if TERM is unset:

    $ TERM= sipp some.host
    Error opening terminal: unknown.

I think should retry screen_init with TERM=dummy if it fails. That way you can start a sipp command from a cron job / zabbix job / whatever without having to resort to -bg mode.

(That takes care of this addition https://code.osso.nl/projects/sipp/changeset/48b81a8bd29f/ )

wdoekes:
(fed347b was superseded by c2381c8 -- fixed log message)

PR #130, opened 2015-05-06
No curses!

vodik:
This patches SIPp to conditionally only enables ncurses if and only if stdout is a tty. This means sipp will now work with its standard out redirected to file, no terminal corruption because of ncurses mode being initialized anyways.

In order to support this, however, the signal handling has to be decoupled from the ncurses support. Hence a slight bit of rewriting and migration of a lot of logic in `screen.cpp` into `sipp.cpp`.

### DIFF
--- a/include/screen.hpp
+++ b/include/screen.hpp
@@ -48,10 +48,10 @@ extern "C" {
 #define EXIT_BIND_ERROR            -2
 
 void screen_set_exename(char * exe_name);
-void screen_init(void (*exit_handler)());
+void screen_init();
 void screen_clear();
 int  screen_readkey();
-void screen_exit(int rc);
-void screen_sigusr1(int /* not used */);
+void screen_exit();
+void screen_show_errors();
 
 #endif // __SCREEN_H__
--- a/include/sipp.hpp
+++ b/include/sipp.hpp
@@ -479,6 +479,8 @@ enum E_Alter_YesNo {
 
 #include "strings.hpp"
 
+void sipp_exit(int rc);
+
 char *get_peer_addr(char *);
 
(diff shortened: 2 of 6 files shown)


wdoekes:
Ah, good. That should fix both #96 and #75.

I'll look into this a bit better later. Thanks for the work!

In the mean time, I do see some code style violations (missing spaces around operators, missing space after keywords (if, while), missing braces and enters after single-line ifs). If you could take care of those, that would make the review easier.

vodik:
Is there a standard guide somewhere? Yeah, I realized there where violations when I dove in there, but the original code in that area had no standardization anywhere.

vodik:
Okay, I cleaned it up a fair bit. The initial patch tried to be as self-contained as possible since the original code base was just soo poorly formatted and confusing. I took some time to up the quality and dropping that dangerous function pointer cast as well.

vodik:
Actually, let me do some more work on this. I should move the `EXIT_*`, the `ERROR`/`WARNING`, etc and the sigusr1/2 logic.

vodik:
Actually looks like those methods are still coupled and what was initially throwing me off: `screen_sigusr1`, is actually an unused definition.

vodik:
Okay, as far as I know, this should be complete breaking no functionality. We've  been using this patch set internally for some time but we have some different logging code (we write the screen contents out to different files in a log directory) that I'm hoping to also put up for consideration.

wdoekes:
(fed347b was superseded by c2381c8 -- fixed log message)

Thanks for the patch!

vodik:
Oh shit, sorry.

Issue #34, opened 2013-10-08
Message body has additional \r\n\r\n in message


rkday:
(Originally https://sourceforge.net/p/sipp/bugs/144/)

When trying to send hex representation of DTMF in NOTIFY message body ( CISCO proprietary) , an additional 4 bytes are added to the message body 0x0a,0x0d,0x0a,0x0d.

e.g.

<![CDATA[
  NOTIFY [next_url] SIP/2.0
  Via: SIP/2.0/[transport] [local_ip]:[local_port];branch=[branch]
  Record-[routes]
  To:[$From_Content]
  From:[$To_Content]
  [last_Call-ID:]
  CSeq: 5 NOTIFY
  Contact: sip:sipp@[local_ip]:[local_port]
  Max-Forwards: 70
  Event:telephone-event
  Content-Type: audio/telephone-event
  Content-Length: 4

  \x0b\x00\x00\x28]]>

PR #133, opened 2015-05-14
A stab at fixing #34: don't add CRLF to just any CDATA

wdoekes:
Not sure why the removed_crlf was only use in the `<send>` code and not the `<sendCmd>`.

Rob, do you see anything wrong with this?

### DIFF
--- a/src/scenario.cpp
+++ b/src/scenario.cpp
@@ -599,14 +599,17 @@ int get_cr_number(const char *src)
     return res;
 }
 
-static char* clean_cdata(char *ptr, int *removed_crlf = NULL)
+static char* clean_cdata(char *ptr)
 {
-    char * msg;
+    char* msg;
+    int removed_crlf = 0;
 
-    while((*ptr == ' ') || (*ptr == '\t') || (*ptr == '\n')) ptr++;
+    while (*ptr == ' ' || *ptr == '\t' || *ptr == '\n') {
+        ptr++;
+    }
 
-    msg = (char *) malloc(strlen(ptr) + 3);
-    if(!msg) {
+    msg = (char*)malloc(strlen(ptr) + 3);
+    if (!msg) {
         ERROR("Memory Overflow");
     }
     strcpy(msg, ptr);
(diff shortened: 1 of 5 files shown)


wdoekes:
@rkday and perhaps we should give a particular someone collaborator powers?

src/scenario.cpp
```diff
@@ -647,6 +644,15 @@ char *clean_cdata(char *ptr, int *removed_crlf = NULL)
         memmove(ptr, ptr + 1, strlen(ptr));
     }
 
+    /* Re-add our removed (CR)LFs. */
+    while (removed_crlf--) {
+        strcat(msg, "\n");
+    }
+    /* A SIP message must contain a double (CR)LF to signify EOH. */
```
rkday: I think it must actually contain a double CRLF - the CR is mandatory. Worth a comment to say that the \r gets added by the SendingMessage constructor so we don't need it here.

rkday:
Looks good, maybe worth a test with:

```
<send>
  <![CDATA[
      SIP/2.0 200 OK
      [last_Via:]
      [last_From:]
      [last_To:];tag=TAG
      [last_Call-ID:]
      [last_CSeq:]
      Contact: sip:sipp@127.0.0.1:5070
      Max-Forwards: 70]]>
  </send>
```

to ensure that EOH gets added even if it's not there.

src/scenario.cpp
wdoekes: Yeah. It's mandatory in SIP. It was an attempt to document that the CR wasn't used here. But you're right: it's better to be explicit (ooh, zen of python).

wdoekes:
Okay. Added the relevant checks.

Now I'm going nuts though, since it seems already fixed, even with an ancient (Ubuntu supplied sip-teste) sipp 1:3.2-1build1. I have absolutely no idea why I cannot reproduce the failing test anymore :-/

wdoekes:
Okay. This changes the newly commited (6a09339) github-#0034 patch to fail:
```
Testing github-#0034: failed (on notify-manylf)
```
The cleanup is not worth the change in behaviour.

PR #134, opened 2015-05-14
Improve SDP parsing to accept \n as line terminator

vodik:
RFC 4566 says parers SHOULD accept both single CR and LF characters as line terminators:

> The sequence CRLF (0x0d0a) is used to end a record, although parsers SHOULD be tolerant and also accept records terminated with a single newline character.

### DIFF
--- a/src/call.cpp
+++ b/src/call.cpp
@@ -148,7 +148,7 @@ uint32_t get_remote_ip_media(char *msg)
         /* Can't find what we're looking at -> return no address */
         return INADDR_NONE;
     }
-    begin += sizeof("c=IN IP4 ") - 1;
+    begin += sizeof(pattern) - 1;
     end = find_sdp_eol(begin);
     if (!end) {
         free(my_msg);
@@ -185,7 +185,7 @@ uint8_t get_remote_ipv6_media(char *msg, struct in6_addr *addr)
         /* Can't find what we're looking at -> return no address */
         return 0;
     }
-    begin += sizeof("c=IN IP6 ") - 1;
+    begin += sizeof(pattern) - 1;
     end = find_sdp_eol(begin);
     if (!end) {
         free(my_msg);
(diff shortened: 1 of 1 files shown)


rkday:
Looks fine to me. I prefer `char* x` to `char *x`, though - the file's already inconsistent between the two, but we should standardise on one rather than staying inconsistent.

vodik:
For what its worth, I prefer `char *x` to `char* x` because the `*` is bound to the variable, not the type. I always thought the alternative style is a mistake. After all: its: 

~~~c
char *foo = NULL, *bar = NULL;
~~~

though whether or not you think defining multiple variables on one line is a good thing or not is another style issue that can be debated.

Anyways, now that I've stated my preference, don't think I'm willing to change since I'm not the one doing the maintenance.

wdoekes:
> don't think I'm willing

I assume you meant "don't think I'm not willing" ;)

Anyways, I've automated the use of your style too (`char *x`), until I do a cast at which point the extra space distracts (`(char *)x`). If I look in the C header files, I see `char *x` everywhere, but if I look in the C++ headers (`/usr/include/c++/4.8/istream`) I see `char* x`. That happens to be Bjarne's stand:
http://www.stroustrup.com/bs_faq2.html#whitespace

> A "typical C++ programmer" writes `int* p;` and explains it "p is a pointer to an int" emphasizing 
type. Indeed the type of `p` is `int*`. I clearly prefer that emphasis and see it as important for using the more advanced parts of C++ well.

So, I'll agree with Rob that we should move to `char* x` for new code.

vodik:
Two against one, oh well...

> I assume you meant "don't think I'm not willing" ;)

Gah, yeah, that's what I meant. I'll fix this up when I got some time later today.

vodik:
Alright, updated.

wdoekes:
Thanks!

PR #138, opened 2015-05-20
Fix potential double free after sendCmdBuffer

vodik:
The stupid `delete this;` error handling insider sendCmdBuffer sucks. We need to bail anytime that function returns a failure or we trigger either double frees or use-after-free bugs.

Caught by clang-analyzer.

I might revisit this to see if I can tackle the memory allocation madness since this isn't really what I'd consider maintainable.


wdoekes:
Oh, I had to look at that several times. sendCmdBuffer really does contain a `delete this;`. That indeed does suck horribly.

PR #140, opened 2015-05-21
Fixes addressing -Wmissing-prototypes and -Wformat=1, few other misc changes.

vodik:
The `quoted_strchr` is a small patch from the Sangoma branch.

Attach the `format` attribute on all the printf-like logging functions. This didn't reveal any problems but good to have. Figured its okay since we're already using `noreturn`.

The rest of the diff tackles correcting all -Wmissing-prototypes warnings. In most cases this means making functions static, in other cases its because the relevant header file isn't being properly included. It also meant fixing argumentless functions in C to correctly use void.


include/rijndael.h
```diff
@@ -18,9 +18,10 @@
 #ifndef RIJNDAEL_H
 #define RIJNDAEL_H
 
+typedef unsigned char   u8;
+typedef unsigned int    u32;
```
wdoekes: Unsigned char might be 8 bits, but unsigned int will not be 32 bits on all platforms.

You'll need better typedefs, stdint.h should provide a nice `uint32_t`.

include/logger.hpp
```diff
@@ -87,21 +87,13 @@ void rotate_calldebugf();
...
-    int LOG_MSG(const char *fmt, ...);
-    /*
-#ifdef __cplusplus
-}
-#endif
-*/
+int TRACE_MSG(const char* fmt, ...) __attribute__((format(printf, 1, 0)));
+int TRACE_CALLDEBUG(const char* fmt, ...) __attribute__((format(printf, 1, 0)));
+int TRACE_SHORTMSG(const char* fmt, ...) __attribute__((format(printf, 1, 0)));
+int LOG_MSG(const char* fmt, ...) __attribute__((format(printf, 1, 0)));
```
wdoekes: Shouldn't that be `__attribute__((format(printf, 1, 2)))` ?

include/screen.hpp
```diff
@@ -31,10 +31,10 @@
...
 extern "C" {
 #endif
-    void ERROR(const char *fmt, ...) __attribute__ ((noreturn));
-    void WARNING(const char *fmt, ...);
-    void ERROR_NO(const char *fmt, ...) __attribute__ ((noreturn));
-    void WARNING_NO(const char *fmt, ...);
+    void ERROR(const char *fmt, ...) __attribute__ ((format(printf, 1, 0), noreturn));
+    void WARNING(const char *fmt, ...) __attribute__((format(printf, 1, 0)));
+    void ERROR_NO(const char *fmt, ...) __attribute__ ((format(printf, 1, 0), noreturn));
+    void WARNING_NO(const char *fmt, ...) __attribute__((format(printf, 1, 0)));
```
wdoekes: also `1, 2`

include/call.hpp
```diff
@@ -322,7 +322,7 @@ class call : virtual public task, virtual public listener, public virtual socket
     BIO *m_bio;
 #endif
 
-    int _callDebug(const char *fmt, ...);
+    int _callDebug(const char *fmt, ...) __attribute__((format(printf, 2, 0)));
```
wdoekes: Should this be `2, 3` because of the hidden this pointer? Or should it be `1, 2`?

src/auth.cpp
```diff
@@ -104,7 +104,7 @@ int createAuthHeaderAKAv1MD5(const char *user,
 
 /* This function is from RFC 2617 Section 5 */
 
-void hashToHex (md5_byte_t *_b_raw, unsigned char *_h)
+static void hashToHex (md5_byte_t *_b_raw, unsigned char *_h)
```
wdoekes: While you're here, you may as well remove the space between the function name and the parenthesis.

src/message.cpp
```diff
@@ -93,6 +93,19 @@ struct KeywordMap SimpleKeywords[] = {
 
 #define KEYWORD_SIZE 256
 
+static char* quoted_strchr(char* s, int c)
+{
+    char* p;
+
+    for (p = s; *p && *p != c; p++) {
+        if (*p == '\"') {
```
wdoekes: Double quotes in single quotes don't need backslashes.

wdoekes:
Looks mostly good. Might as well fix the pointers `*` positions in the lines (of C++ files) you're touching, while you're at it.

I'm a bit reluctant to add the __attributes__ (non-portable), but seeing that they're already used, we should probably #define them away on unknown/unsupporting compilers at a later stage.

src/message.cpp
```diff
@@ -93,6 +93,19 @@ struct KeywordMap SimpleKeywords[] = {
 
 #define KEYWORD_SIZE 256
 
+static char* quoted_strchr(char* s, int c)
+{
+    char* p;
```
wdoekes: `const char* s` and `const char* p`, we're not modifying anything.

include/rijndael.h
vodik: I have no problem fixing this stuff up, I was thinking the same thing when I was exploring the code, but this is just me moving around what's already there, just to be clear. Fixes should come in a different commit.

include/logger.hpp
vodik: From the docs:

>  For functions where the arguments are not available to be checked (such as vprintf), specify the third parameter as zero.

Yeah, I got that backwards. Good catch.

include/call.hpp
vodik: No, there's an implicit 'this' argument that gcc counts. I thought so too but got a nice warning.

src/message.cpp
vodik: Can't, since the function consuming is expects `char*`. And that can't be fixed since it modified the string (adds `\0`s).

include/call.hpp
vodik: Yeah, in case you're curious:

~~~
./call.hpp:325:57: error: format attribute cannot specify the implicit this argument as the format string
    int _callDebug(const char *fmt, ...) __attribute__((format(printf, 1, 0)));


~~~

vodik:
Good catch on the format attribute problem. I thought it was fishy there were no formatting mistakes. I additionally replaces `u8`/`u32` with the correct stdint stuff across the entire project since you brought that up.

wdoekes:
`Simplify E_Message_Server_IP ...` shouldn't be in this pull request.

vodik:
Oh shit. Dropped.

vodik:
Okay!

Added `-Wall -pedantic`, fixed up even more of the resulting warning, and snuck in C/C++ standard and attached the getnameinfo tweak. I won't touch this anymore to prevent this from growing into an unmanageable changeset.

God damn whitespace!

We're down to one `-Wvla` and `-Wvariadic-macros` because those aren't supported in c++ and a few `-Wunused-function` that show up when you don't build with pcap play support. A few more warning crop up when those features are compiled in but I don't feel like making this grow any bigger until it lands.

rkday:
Looks good to me - I have fixes for the -Wvla and -Wvariadic-macros warnings, but probably better to merge this and then get those fixes in separately.

@wdoekes - it looks like all your comments have been addressed, but any further thoughts before I merge?

wdoekes:
Awesome! 

Manually merged by be9ad4b.

PR #142, opened 2015-05-27
Disable SDP parsing when you use ignoresdp="true".

wdoekes:
Suggested by @kovacsbalu. This will allow you to keep your media stream
going while rejecting the SDP from a reINVITE.

This obsoletes #97.

I did not test this, but it has better code style and better documentation than the original pull request.

I'm not entirely sure about:
- adding more booleans to the message class
- the naming of ignoresdp

### DIFF
--- a/include/scenario.hpp
+++ b/include/scenario.hpp
@@ -105,6 +105,7 @@ public:
     int            counter;
     double         lost;
     int            crlf;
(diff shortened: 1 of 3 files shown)


kovacsbalu:
Hi @wdoekes, what about this PR?

Issue #137, opened 2015-05-20
sipp v3.4.1 memory leak issue


gomting30:
When we started ipv6 load test with sipp v3.4.1, there was critical memory leak issue at UAS side. sipp process only is using up  more than gigabyte of memory. I don't know what is the problem. it might be the problem of sipp v3.4.1 itself or UAS scenario problemn but i am not sure.

1. uac command line:

        ../sipp -r 300 -rp 1000 -inf usr.csv -sf uac.xml \
          -i [fd00:0:0:7ea1:10:1:1:80] -trace_err [fd00:0:0:7ea1:10:1:1:91]:5060

2. uas command line:

        ../sipp -sn uas -trace_err -sf mrf.xml -p 5010 \
          -i [fd00:0:0:7ea1:10:1:1:80]

3. usr.csv file

        RANDOM,PRINTF=300
        01088880%03d;01099990%03d

4.uas scenario file
        
    <?xml version="1.0" encoding="ISO-8859-1" ?>
    <!DOCTYPE scenario SYSTEM "sipp.dtd">
    
    
    
    
    
    
    
    
    
    
    
    
    
    
    
    
    
    
    
    <scenario name="uas_ct">
      
      
      
      <recv request="INVITE">
        
    
      </recv>
    
      
      
      
      
      
      
      
      
      
      
    
      
    
    
      <send>
        <![CDATA[
    
          SIP/2.0 100 Trying
          [last_Via:]
          [last_From:]
          [last_To:];tag=1234567890_[call_number]
          [last_Call-ID:]
          [last_CSeq:]
          Content-Length: 0
    
        ]]>
      </send>
    
      
    
       
       <send>
        <![CDATA[
    
          SIP/2.0 200 OK
          [last_Via:]
          [last_From:]
          [last_To:];tag=1234567890_[call_number]
          [last_Call-ID:]
          [last_CSeq:]
          Allow: INVITE, ACK, BYE, OPTIONS, CANCEL, INFO, PRACK, UPDATE
          Accept: application/sdp, text/*, application/msml+xml, application/moml+xml
          Contact: <sip:msml@[local_ip]:[local_port];transport=[transport]>
          Content-Length: [len]
          Content-Type: application/sdp
    
          v=0
          o=user1 53655765 2353687637 IN IP[local_ip_type] [local_ip]
          s=-
          c=IN IP[media_ip_type] [media_ip]
          t=0 0
          m=audio [media_port] RTP/AVP 0 96
[...]

PR #144, opened 2015-05-29
Simplify E_Message_Server_IP logic with getnameinfo

vodik:
Folds both code paths into one by using getnameinfo. Removes the malloc'd memory with is never freed, thus closing #137.


wdoekes:
I don't mind this fix. Does it do the same in less lines?

vodik:
Yeah it does, and I'll try and get it in. The more AF independent code the better. I really just closed this because as presented its unnecessary. I'll tackle this tomorrow, its getting really late here.

PR #145, opened 2015-05-31
Change pcap search logic so we prefer files found next to the scenario file.

vodik:
One of the breaking changes with the Sangoma SIPp fork is that we changed the pcap search logic to prefer looking next to the scenario file. This made organizing a large collection of different sipp scenarios considerably simpler.

This is a more thorough patch to the one we have internally, refactoring some other logic as well. I was going to initially prefer checking the current working directory and then, failing finding a pcap there, search next to the scenario. While discussing this change with @wdoekes, he suggested that the current behaviour could actually be considered a bug, which prompted to instead do the reverse, prefer the file next to the scenario instead. This probably makes way more sense and what would be expected anyways.

So, assuming we go with this approach, this patch also adds some additional warning messages to help with the migration:

- Warn if the pcap can't be found next to the scenario that we'll fallback to assuming the file is in the current working directory
- Warn if the pcap happens to be found in both places that now the one next to the scenario will be used.

Catch: the second warning can have false positives: when we provide a full path to the scenario which also happens to be in the current working directory as well. Either I should check if `scenario_path == getcwd()` or just drop the second warning. I'm not convinced its a common enough occurrence.

### DIFF
--- a/src/send_packets.c
+++ b/src/send_packets.c
@@ -53,8 +53,6 @@
 #include "prepare_pcap.h"
 #include "screen.hpp"
 
-#define MAX_PATH 250
-
 extern char* scenario_path;
 extern volatile unsigned long rtp_pckts_pcap;
 extern volatile unsigned long rtp_bytes_pcap;
--- a/src/send_packets.c
+++ b/src/send_packets.c
@@ -53,6 +53,9 @@
 #include "prepare_pcap.h"
 #include "screen.hpp"
 
+#define MAX_PATH 250
+
+extern char* scenario_path;
 extern volatile unsigned long rtp_pckts_pcap;
 extern volatile unsigned long rtp_bytes_pcap;
 extern int media_ip_is_ipv6;
@@ -86,11 +89,29 @@ float2timer (float time, struct timeval *tvp)
     tvp->tv_usec = n * 100000;
 }
 
-/* buffer should be "file_name" */
-int
-parse_play_args (char *buffer, pcap_pkts *pkts)
+static char* find_file(const char* filename)
 {
-    pkts->file = strdup (buffer);
+    if (filename[0] == '/' || !scenario_path) {
+        return strdup(filename);
+    }
+
+    char *fullpath = malloc(MAX_PATH);
+    snprintf(fullpath, MAX_PATH, "%s/%s", scenario_path, filename);
+
+    if (access(fullpath, R_OK) < 0) {
+        free(fullpath);
+        WARNING("SIPp now prefers looking for pcap files next to the scenario. "
+                "%s couldn't be found next to the scenario, falling back to "
+                "using the current working directory", filename);
+        return strdup(filename);
+    }
+
+    return fullpath;
+}
+
+int parse_play_args(const char* filename, pcap_pkts* pkts)
+{
+    pkts->file = find_file(filename);
     prepare_pkts(pkts->file, pkts);
     return 1;
 }
--- a/src/sipp.cpp
+++ b/src/sipp.cpp
@@ -1256,25 +1256,24 @@ static void sighandle_set()
     sigaction(SIGXFSZ, &action_file_size_exceeded, NULL);  // avoid core dump if the max file size is exceeded
 }
 
-static char* remove_pattern(char* P_buffer, char* P_extensionPattern)
+static void set_scenario(const char* name)
 {
+    free(scenario_file);
+    free(scenario_path);
 
-    char *L_ptr = P_buffer;
-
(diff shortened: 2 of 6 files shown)


src/sipp.cpp
```diff
@@ -2271,10 +2267,4 @@ int main(int argc, char *argv[])
     close(epollfd);
     free(epollevents);
 #endif
-
-    if (scenario_file != NULL) {
-        delete [] scenario_file;
-        scenario_file = NULL;
-    }
-
```
vodik: I'm not sure what you guys think of this. IMHO, its pointless to waste time here, the OS will just reclaim that memory anyways...

Unless we care about valgrind, but we can always provide a suppression file.

wdoekes: I do care about valgrind. If I want to find leaks, it doesn't help that we're leaking on purpose. Even though it's slightly faster to let just the OS clean it up, I find it poor coding practise. (In fact, I did the exact opposite in the fix/valgrind branch 6076caa.)

- Maintaining a suppression file is about as much work as simply cleaning up your memory.
- A suppression file probably only works for valgrind and not for other modes of memory leak detection.
- Much of the stuff *is* cleaned up right now. If we decide to ignore some, that's just sloppy. There should at least be some way to identify that a particular variable or set of variables won't be cleaned up.

src/send_packets.c
```diff
@@ -93,11 +94,37 @@ float2timer (float time, struct timeval *tvp)
...
-parse_play_args (char *buffer, pcap_pkts *pkts)
+static char* find_file(const char* filename)
 {
-    pkts->file = strdup (buffer);
+    if (filename[0] == '/' || !scenario_path) {
+        return strdup(filename);
+    }
+
+    char* fullpath = NULL;
+    if (asprintf(&fullpath, "%s/%s", scenario_path, filename) < 0) {
```
wdoekes: GNU extension.

src/send_packets.c
```diff
@@ -93,11 +94,37 @@ float2timer (float time, struct timeval *tvp)
...
+        ERROR_NO("Failed to allocate filepath");
+    }
+
+    if (access(fullpath, R_OK) < 0) {
+        free(fullpath);
+        WARNING("SIPp now prefers looking for pcap files next to the scenario. "
+                "{} couldn't be found next to the secnario, falling back to "
```
wdoekes: Since when does vprintf take python-format-style `{}` parameters? (Shouldn't this have thrown a `-Wformat-extra-args` error?)

And "secnario".

vodik: This is what I get for coding while half asleep...

src/sipp.cpp
vodik: Alright, I won't undo that work. I saw the mess the 3.2 was first hand.

src/send_packets.c
vodik: Yeah, I don't consider this a final version. Technically we're already relying on other GNU extensions, so I thought it be okay for a review. Two things I wanted to bring up.

- `MAX_PATH` is broken everywhere but Windows. And SIPp hardcodes an even smaller maximum that the platforms it supports (ie: for linux its 4098, the code hardcodes 512).
- `asprintf` is useful and easy to implement for platforms that don't have it (call `snprintf` twice).

But yeah, I can just use `MAX_PATH` here like everywhere else in the code base and tackle that issue all at once separately.

vodik: Alright, I think I'll just change this so `MAX_PATH` based logic and open another issue about this that can be perused separately.

vodik:
I dropped the second warning because of the large possibility of false positives. I bet in most cases people are running there scenarios from the current working directory anyways. I'm going to raise a separate issue for `MAX_PATH` when this lands.

vodik:
This should be more in line with the legacy behaviour.

src/send_packets.c
```diff
@@ -93,11 +94,29 @@ float2timer (float time, struct timeval *tvp)
...
+    char *fullpath = malloc(MAX_PATH);
+    snprintf(fullpath, MAX_PATH, "%s/%s", scenario_path, filename);
+
+    if (access(fullpath, R_OK) < 0) {
+        free(fullpath);
+        WARNING("SIPp now prefers looking for pcap files next to the scenario. "
+                "%s couldn't be found next to the scenario, falling back to "
+                "using the current working directory", filename);
+        return strdup(filename);
+    };
```
wdoekes: Stray semi.

wdoekes:
Nice. LGTM. Could you rebase so I can merge?

vodik:
This should do it.

PR #146, opened 2015-05-31
Automate version info using git describe

vodik:
This pulls out the `SIPP_VERSION` macro to favour of the `VERSION` macro provided by `config.h`. This is something we probably want in general so we can do away with the "don't forget to update configure.ac" because we really don't have to remember to update both.

The `m4_esyscmd_s([git describe --tags --always --dirty])` trick lets us use git to provide the version info. After compiling sipp with this change set, I now get the much more descriptive:

```
$ sipp -v
 SIPp v3.5.0-rc1-52-ga3a53de-dirty-TLS-SCTP-PCAP-RTPSTREAM built May 30 2015, 23:44:00.
```

Two things to note:

1. `--dirty` might not actually be desirable, as I'm pretty sure all builds will end up dirty. Building sipp marks the gtest submodule as dirty and thus the sipp parent repo. The first build should be clean but subsequent builds will be dirty.
2. `--tags` is there because versions have been marked with regular tags. You guys should seriously consider switching to annotated tags for versioning. At the very least you'll leave users with a record of *who* made a release (and they can be signed).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,8 @@
 # ======================== initialization ===============================
 
-AC_INIT([SIPp], [3.5.0-rc1], [sipp-users@lists.sourceforge.net], [sipp])
+AC_INIT([SIPp],
+        [m4_esyscmd_s([git describe --tags --always])],
+        [sipp-users@lists.sourceforge.net], [sipp])
 
 # Load non-standard m4 files from ./m4/*.m4.
 # AC_CONFIG_MACRO_DIR([m4]) does not work on pre-1.13 aclocal, so we
--- a/include/sipp.hpp
+++ b/include/sipp.hpp
@@ -104,8 +104,6 @@
 
 /************************** Constants **************************/
 
-#define SIPP_VERSION               "3.5.0-rc1" /* don't forget to update configure.ac */
-
 #define T_UDP                      0
 #define T_TCP                      1
 #define T_TLS                      2
--- a/src/call.cpp
+++ b/src/call.cpp
@@ -53,6 +53,7 @@
 #include "sipp.hpp"
 #include "auth.hpp"
 #include "deadcall.hpp"
+#include "config.h"
 
 #define callDebug(args...) do { if (useCallDebugf) { _callDebug( args ); } } while (0)
 
@@ -2314,7 +2315,7 @@ char* call::createSendingMessage(SendingMessage *src, int P_index, char *msg_buf
             dest += snprintf(dest, left, "%d", userId);
             break;
         case E_Message_SippVersion:
-            dest += snprintf(dest, left, "%s", SIPP_VERSION);
+            dest += snprintf(dest, left, "%s", VERSION);
             break;
         case E_Message_Variable: {
             int varId = comp->varId;
--- a/src/sipp.cpp
+++ b/src/sipp.cpp
@@ -44,6 +44,7 @@
 #include "socket.hpp"
 #include "logger.hpp"
 #include "assert.h"
+#include "config.h"
 
 extern struct sipp_socket *ctrl_socket;
 extern struct sipp_socket *stdin_socket;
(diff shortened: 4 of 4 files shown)


vodik:
Oh, the other thing worth bringing up is that `E_Message_SippVersion` will now have the 'v' prepended to the message. I'm not sure that matters. If it does it can be compensated for.

configure.ac
```diff
@@ -1,6 +1,8 @@
 # ======================== initialization ===============================
 
-AC_INIT([SIPp], [3.5.0-rc1], [sipp-users@lists.sourceforge.net], [sipp])
+AC_INIT([SIPp],
+        [m4_esyscmd_s([git describe --tags --always --dirty])],
```
wdoekes: If the version is "hardcoded" in the Makefile, we should force a reconfigure by make if the git version changes; lest we forget to reconfigure before making a release

The drawback would be that you get a reconfigure after every local commit: no fun.

I'd rather see only a version.c file rebuilt upon git-version-change.

wdoekes:
P.S. I don't mind the changed `E_Message_SippVersion`.

wdoekes:
P.P.S. I don't thing `--dirty` adds anything beneficial either. And +1 to annotated tags.

configure.ac
vodik: Yeah, but this isn't for that. This is meant for helping releases/debugging people who build from git:

1. We release a tarball. To do that sanely, we should run the first two steps of `build.sh`: `git submodule update --init` and `autoreconf -vifs` and tar up the result (without `.git` included obviously). The resulting `./configure` script is now setup with a version pulled from git describe. No manual tweaking of code before tagging a release.
2. If someone builds straight from git (like if I was to package a -git version of sipp internally), the `./sipp -v` will be reflective of HEAD when it was built.

I don't really see this as useful for manual in-tree work since the usefulness of this goes away if the tree is dirty anyways.

rkday:
Using the autoconf VERSION instead of a separate-but-identical SIPP_VERSION looks good to me (that's just sensible de-duplication), but I'm a bit nervous about having the autoconf version just come from `git describe --tags` - I think that means you can't use it with a tarball from https://github.com/SIPp/sipp/releases, which is a shame.

I think I'd rather see "3.5.0-rc1" in the AC_INIT, but to also (if `git describe` gives sensible output) give the git commit ID separately in the `-v` output. Something like https://github.com/SIPp/sipp/commit/d264a76bc3470163544d6b2cd1b40cc8675a3653?

vodik:
> I think that means you can't use it with a tarball from https://github.com/SIPp/sipp/releases, which is a shame.

Well that's already the case. Its really broken. Download it, the gtest folder is empty. And since its not a git repo, the build script breaks: 

    fatal: Not a git repository (or any parent up to mount point /tmp)
    Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).

Removing the submodule update line doesn't help much because we force the unit testing: `src/actions.cpp:748:10: fatal error: 'gtest/gtest.h' file not found`.

Technically, as far as I know for best autoconf practices, git's tarball support isn't ideal anyways. My understanding of them is that you don't want any of the auto-generated files for development, but you want them in your release. That way users can just do `tar xfvz ...; ./configure; make; make install` to build, no messing with autoconf. And if we support that, then this mechanism will mean we always ship the correct version without fuss. Tag then make tarball, no code change.

Now that's only a best practise. We can be smarter with the build.sh script and we can certainly deviate. It's a shame git doesn't allow uploading custom tarballs anymore.

vodik:
I'll break this up into two commits in the meanwhile so at least the macro deduplication can be pulled in.

wdoekes:
b2fcf33 was cherry-picked and then amended (re-added the missing "v").

wdoekes:
#153 works like I'd expect it to.

You can store a `.version` in  tarball to set the right version there.

PR #147, opened 2015-06-04
_BSD_SOURCE macro has been deprecated

vodik:


wdoekes:
> To allow code that requires
> _BSD_SOURCE in glibc 2.19 and earlier and _DEFAULT_SOURCE in
> glibc 2.20 and later to compile without warnings, define both
> _BSD_SOURCE and _DEFAULT_SOURCE.

Even my pretty up to date Ubuntu LTS (14) uses 2.19, so removing BSD_SOURCE altogether is too early. Better add it back next to _DEFAULT_SOURCE.

```
$ dpkg -l libc6 | sed -ne6p
ii  libc6:amd64                                           2.19-0ubuntu6.6
```

And adding a little reference to the deprecation docs in the commit msg doesn't hurt:

> feature_test_macros(7) wrote:
> ...

vodik:
Oh shit, I didn't realize that was a recent thing. Will research what's the right way to support both.

vodik:
Looking in the `feature_test_macros(7)` page, looks like this is the default behaviour, at least one linux: 

> If no feature test macros are explicitly defined, then the following feature test macros are defined by default: _BSD_SOURCE (in glibc 2.19 and earlier), _SVID_SOURCE (in glibc 2.19 and earlier), _DEFAULT_SOURCE (since glibc 2.19)...

I wonder what the behaviour is on other platforms we support, because at least on Linux this seems like it could be dropped.

wdoekes:
Thanks :)

PR #148, opened 2015-06-07
Allow per-IP TCP sockets

rkday:
Adds a '-t ti' option that's the TCP analogue of '-t ui'.


src/call.cpp
```diff
@@ -941,6 +941,38 @@ bool call::connect_socket_if_needed()
 
         sipp_customize_socket(call_socket);
 
+        if (peripsocket) {
+            struct sockaddr_storage saddr;
+            char peripaddr[256];
```
vodik: Shouldn't this be `SIPP_MAX_MSG_SIZE`? Or am I mistaken? To my understanding, it looks like `getFieldFromInputFile` assumes that buffer length.

src/call.cpp
```diff
@@ -941,6 +941,38 @@ bool call::connect_socket_if_needed()
...
+            char *tmp = peripaddr;
+            getFieldFromInputFile(ip_file, peripfield, NULL, tmp);
+            struct addrinfo * h ;
+            struct addrinfo   hints;
+            memset((char*)&hints, 0, sizeof(hints));
+            hints.ai_flags  = AI_PASSIVE;
+            hints.ai_family = PF_UNSPEC;
```
vodik: This should be `AF_UNSPEC`. The distinction is kinda moot but we're talking about the address family, not the protocol family.

vodik:
Have to say, looks useful.

src/call.cpp
```diff
@@ -941,6 +941,38 @@ bool call::connect_socket_if_needed()
...
         sipp_customize_socket(call_socket);
 
+        if (peripsocket) {
+            struct sockaddr_storage saddr;
+            char peripaddr[256];
+            char *tmp = peripaddr;
+            getFieldFromInputFile(ip_file, peripfield, NULL, tmp);
+            struct addrinfo * h ;
+            struct addrinfo   hints;
+            memset((char*)&hints, 0, sizeof(hints));
```
wdoekes: Use struct initializer `... hints = {0,}` and drop the memset.

src/call.cpp
```diff
@@ -941,6 +941,38 @@ bool call::connect_socket_if_needed()
 
         sipp_customize_socket(call_socket);
 
+        if (peripsocket) {
+            struct sockaddr_storage saddr;
+            char peripaddr[256];
+            char *tmp = peripaddr;
+            getFieldFromInputFile(ip_file, peripfield, NULL, tmp);
```
wdoekes: What good does the addition of `char *tmp = ...` do? Looks to me like `peripaddr` could simply be passed twice as args 2 and 4..

(And I believe we settled on `char*`.)

src/call.cpp
```diff
@@ -941,6 +941,38 @@ bool call::connect_socket_if_needed()
...
+            struct addrinfo * h ;
+            struct addrinfo   hints;
+            memset((char*)&hints, 0, sizeof(hints));
+            hints.ai_flags  = AI_PASSIVE;
+            hints.ai_family = PF_UNSPEC;
+            getaddrinfo(peripaddr,
+                        NULL,
+                        &hints,
+                        &h);
```
wdoekes: A bit superfluous indent, imo.

src/call.cpp
```diff
@@ -941,6 +941,38 @@ bool call::connect_socket_if_needed()
...
+            if (use_ipv6) {
+                (_RCAST(struct sockaddr_in6 *, &saddr))->sin6_port = htons(0);
+            } else {
+                (_RCAST(struct sockaddr_in *, &saddr))->sin_port = htons(0);
+            }
+            int port = -1;
+            int ret = sipp_bind_socket(call_socket, &saddr, &port);
+             if (ret != 0) {
+              WARNING_NO("Could not bind to %s: %d", peripaddr, ret);
+            }
```
wdoekes: Bad indentation.

src/call.cpp
```diff
@@ -941,6 +941,38 @@ bool call::connect_socket_if_needed()
...
+        }
+
+
```
wdoekes: I'd favor it if the whole getaddrinfo thing got moved out of here. I believe I've seen the same getaddrinfo code in various places.

```
if (peripsocket) {
    struct sockaddr_storage saddr;
    char peripaddr[256];
    getFieldFromInputFile(ip_file, peripfield, NULL, peripfield);
    sipp_aton(use_ipv6, &saddr, peripfield);
    int port = -1;
    int ret = sipp_bind_socket(call_socket, &saddr, &port);
    ...
```

src/sipp.cpp
```diff
@@ -1567,9 +1567,6 @@ int main(int argc, char *argv[])
                     break;
                 }
 
-                if (peripsocket && transport != T_UDP) {
-                    ERROR("You can only use a perip socket with UDP!\n");
-                }
```
wdoekes: Does it work for all transports? Or just for T_UDP and T_TCP ?

src/call.cpp
wdoekes: For the record:

socket(2) says PF_* is deprecated, so AF_* should be preferred anyway.
In line with those docs, e.g. packet(7) uses only AF_PACKET in the manual.

src/call.cpp
vodik: Yeah, talking about refactoring, thus could use some work too. For example, we're passing an empty service field to `getaddrinfo` and then doing AF-dependant magic to inject the desired port.

src/call.cpp
rkday: I've changed getFieldFromInputFile to take a length parameter - we don't need 64kb for an IP address.

src/call.cpp
rkday: It couldn't previously (getFieldFromInputFile took a reference to a char*), but I've refactored to make it better.

src/call.cpp
rkday: I've pulled this logic ("take an IP address char* and fill in a struct sockaddr_storage") out into a separate function, although I haven't used it everywhere yet.

src/call.cpp
vodik: Yeah, even better.

src/call.cpp
```diff
@@ -904,27 +895,13 @@ bool call::connect_socket_if_needed()
...
-                       _RCAST(struct sockaddr_storage *,h->ai_addr)));
-
-            if (use_ipv6) {
-                (_RCAST(struct sockaddr_in6 *, &saddr))->sin6_port = htons(local_port);
-            } else {
-                (_RCAST(struct sockaddr_in *, &saddr))->sin_port = htons(local_port);
-            }
+                    local_addr_storage->ai_addr,
+                    SOCK_ADDR_SIZE(
+                        _RCAST(struct sockaddr_storage *,local_addr_storage->ai_addr)));
```
vodik:
~~~c
memcpy(&saddr, local_addr_storage->ai_addr, sizeof(struct sockaddr_storage));
~~~

Is perfectly safe and simpler.

src/socket.cpp
```diff
@@ -510,13 +510,9 @@ void setup_ctrl_socket()
 
     memset(&ctl_sa, 0, sizeof(struct sockaddr_storage));
     if (control_ip[0]) {
-        struct addrinfo hints;
+        struct addrinfo hints = {AI_PASSIVE, AF_UNSPEC,};
```
vodik: I really wish we had C style struct initialization in C++... :/

src/socket.cpp
```diff
@@ -3076,6 +2963,29 @@ bool sipMsgCheck (const char *P_msg, struct sipp_socket *socket)
     return false;
 }
 
+bool fill_sockaddr_from_ip(struct sockaddr_storage* saddr, char* ip, bool use_ipv6)
+{
+    struct addrinfo* h ;
+    struct addrinfo hints = {AI_PASSIVE, AF_UNSPEC,};
+    if (!getaddrinfo(ip, NULL, &hints, &h))
```
vodik:
~~~c
char service[std::numeric_limits<unsigned short>::digits10 + 1];
snprintf(service, sizeof(service), "%hs", local_port);
if (!getaddrinfo(ip, service, &hints, &h);
//...
~~~

And we don't have to worry about any of the problems now with `(_RCAST(struct sockaddr_in6 *, saddr))`. That said, I'd also consider this a temporary fix, ideally we should keep the port as a number.

wdoekes:
(And don't forget to squash before merging, since the commits aren't logical changesets.)

src/call.cpp
```diff
@@ -855,7 +855,7 @@ bool call::connect_socket_if_needed()
     if(!multisocket) return true;
 
     if(transport == T_UDP) {
-        struct sockaddr_storage saddr;
+        struct sockaddr_storage saddr = {0,};
```
vodik:
~~~c
struct sockaddr_storage saddr = {};
~~~

Is enough in c++11.

src/call.cpp
```diff
@@ -881,50 +879,29 @@ bool call::connect_socket_if_needed()
...
+                    SOCK_ADDR_SIZE(
+                        _RCAST(struct sockaddr_storage *,local_addr_storage->ai_addr)));
+        
             if (use_ipv6) {
-                (_RCAST(struct sockaddr_in6 *, &saddr))->sin6_port = htons(local_port);
+                saddr.ss_family       = AF_INET6;
             } else {
-                (_RCAST(struct sockaddr_in *, &saddr))->sin_port = htons(local_port);
+                saddr.ss_family       = AF_INET;
             }
```
vodik: The setting of `ss_family` shouldn't be necessary, it should always be set.

src/call.cpp
```diff
@@ -881,50 +879,29 @@ bool call::connect_socket_if_needed()
...
+                    local_addr_storage->ai_addr,
+                    SOCK_ADDR_SIZE(
+                        _RCAST(struct sockaddr_storage *,local_addr_storage->ai_addr)));
+        
             if (use_ipv6) {
-                (_RCAST(struct sockaddr_in6 *, &saddr))->sin6_port = htons(local_port);
+                saddr.ss_family       = AF_INET6;
             } else {
-                (_RCAST(struct sockaddr_in *, &saddr))->sin_port = htons(local_port);
+                saddr.ss_family       = AF_INET;
```
vodik: The setting of `ss_family` shouldn't be necessary, it should always be set.

src/call.cpp
```diff
@@ -881,50 +879,29 @@ bool call::connect_socket_if_needed()
...
+            fill_sockaddr_from_ip(&saddr, peripaddr, use_ipv6 ? AF_INET6 : AF_INET);
+        } else {
             memcpy(&saddr,
-                   h->ai_addr,
-                   SOCK_ADDR_SIZE(
-                       _RCAST(struct sockaddr_storage *,h->ai_addr)));
-
+                    local_addr_storage->ai_addr,
+                    SOCK_ADDR_SIZE(
+                        _RCAST(struct sockaddr_storage *,local_addr_storage->ai_addr)));
```
vodik: #149 kills the `SOCK_ADDR_SIZE` macro, and we don't need it here anyways:

```c
memcpy(&saddr, local_addr_storage->ai_addr, local_addr_storage->ai_addrlen);

File: src/socket.cpp [hunk 16]
```diff
@@ -2609,111 +2567,44 @@ int open_connections()
...
+
+        // Bind socket to local_ip
+        if (peripsocket) {
+            // On some machines it fails to bind to the self computed local
+            // IP address.
+            // For the socket per IP mode, bind the main socket to the
+            // first IP address specified in the inject file.
+            inFiles[ip_file]->getField(0, peripfield, peripaddr, sizeof(peripaddr));
+            if (!fill_sockaddr_from_ip(&local_sockaddr, peripaddr, local_ip_is_ipv6 ? AF_INET6 : AF_INET))
+            {
```
vodik: Maybe we want to have a consistent style on brackets?

src/socket.cpp
```diff
@@ -3076,6 +2905,29 @@ bool sipMsgCheck (const char *P_msg, struct sipp_socket *socket)
...
+    if (h->ai_family == AF_INET6) {
+        (_RCAST(struct sockaddr_in6 *, saddr))->sin6_port = htons(local_port);
+        saddr->ss_family       = AF_INET6;
+    } else {
+        (_RCAST(struct sockaddr_in *, saddr))->sin_port = htons(local_port);
+        saddr->ss_family       = AF_INET;
+    }
+    freeaddrinfo(h);
+    return true;
+}
```
vodik: Yeah, if we're copying `ai_addr` into `saddr`, `ss_family` will be set already.

vodik:
Very nice, this really cleans up the code base.

wdoekes:
If this is still useful, it will require a rebase and a squash.

wdoekes:
The merge conflicts are overwhelming. Assuming this is left for dead. Feel free to reopen :)

Issue #109, opened 2014-11-26
Merge with Sangoma's sipp fork


vodik:
Hey, so we have our own internal sipp fork, started off of sipp-3.2 release, which we would like to spend some time merging upstream. We currently don't have it hosted online here yet... but there's quite a bit of work you guys might be interested in. Some stuff I listed may or may not have been done again independantly - I haven't had a chance to fully discover this repo's code base, so sorry if I list what appears a non-feature

- Decoupled sipp from ncurses so sipp behaves correctly when stdout isn't a tty (I want to get to the point where its fully optional)
- Reworked pcap playback: support for more pcap datalinks (only DTL_LINUX_SLL tested atm) and setting a pcap search directory
- Ability to inject additional headers into a sipp script from command line arguments (we wanted something a little more powerful then variable expansion)
- Slightly more logging
- Misc fixes in ipv6 handling (things that I think have been fixed already) and source port/address binding for tcp

vodik:
Alright, so my time has opened up a bit and I'm going to try and get the rest of this pushed through. Hopefully the next set of stuff should be ncurses decoupling. It just a bit painful to backport changes since I'm porting pre-uncrustified diffs :sob:.

Is there any interest in a python/pytest wrapper around sipp? We got one kicking around, I don't know if you guys would be interested in it.

wdoekes:
> support for more pcap datalinks (only DTL_LINUX_SLL tested atm)

Has been added just now in 0026639.

> Is there any interest in a python/pytest wrapper around sipp? We got one kicking around, I don't know if you guys would be interested in it.

Not sure. I already made a few tests with bash in the /regress/ directory. I don't have an opinion about moving to python, but it shouldn't hurt.

vodik:
Well its not tests for sipp, its for using sipp to test something else in an automated way. Scenario testing.

wdoekes:
Then I'm not sure it fits in here. You can always throw it on github and see it interest is sparked.

I bet there are a lot of python (and ruby, and bash) test suites that incorporate calls to one or several sipp instances; all with their own advantages and disadvantages.

vodik:
Fair enough

vodik:
Okay, so here's the status:

- Most of the bugs I found have been already fixed here. Namely handling ipv6 addresses segfaulting due to poor parsing and tcp transport not binding to the `user_port` under some conditions. SDP parsing was the only issue I found still broken (and its been fixed).
[...]

PR #149, opened 2015-06-08
Some cleanups for input on (not necessarily for merging)

vodik:
I thought I'd put this up as a merge request mainly for the peer review, I'm not really expecting this to get merged as is, though probably safe to do so.

#145 is the last merge request that I'd consider sufficient to land for issue #109 to be closed. So going forward I want to start putting effort into cleaning up the code base. This change set, thus far:

1. Cleans up `actions.cpp` code to line up with the set standard.
2. Removes some dead functions and types - stuff that's not used anywhere.
3. Removes trivial getters and setters. There's no point, the member could just be exposed publicly.
4. Defines a copy and move constructor for `CAction` (afaik, the default implementations are fine).
5. Removes `CActions` completely and replaces it with a `vector`.

And a few other misc changes. Dropping `CActions` for a vector might improve cache locality and thus performance, but I don't have any benchmarks to back that up.

If you guys are amicable to this kind of work, I plan to keep it coming. But I'm bringing it up to get feedback and an idea of how comfortable you guys are with these kinds of changes.


vodik:
Also what are your guys' opinions on the prefixing? eg `CAction`, `T_IntCmdType`, `M_scenario`, which are inconsistently done only in parts of the code base and imho don't add anything of value to the names.

wdoekes:
> Also what are your guys' opinions on the prefixing?

The only prefix I find somewhat useful is an `m_` prefix for member attributes. (Particularly in this codebase where many globals are used as well; "is this a local, member or global var?") But I can certainly live without them.

vodik:
I just realized this is growing quite large, I'll stop pushing dev work to this branch.

wdoekes:
Awesome cleanup work!

Code changes look good. Compiles (after removing noexcept). Regression tests run.

And I like this bit a lot:
```
$ git diff | wdiffcount 
1367 - 1792 = -425
```

vodik:
Yeah and a lot less manual memory management. I think I've addressed everything.

vodik:
Okay, so  this was prematurely merged, I'm going to the mature thing and yank it. Sorry, it should have been more extensively tested and I'm not confident I've fixed everything with the next damage control merge.

I'll polish this one up more in private before posting it again. Sorry again.

Issue #150, opened 2015-06-19
EAFNOSUPPORT on ipv6 regression


vodik:
The following command causes `EAFNOSUPPORT` errors in `./sipp -sn uac '[::1]:12347' -p 12348`. Looking at an strace, looks like:

``` strace
socket(PF_INET, SOCK_DGRAM, IPPROTO_UDP) = 4
bind(4, {sa_family=AF_INET, sin_port=htons(12348), sin_addr=inet_addr("0.0.0.0")}, 16) = 0
sendto(4, "INVITE sip:service@[::1]:12347 S"..., 512, 0, {sa_family=AF_INET6, sin6_port=htons(12347), inet_pton(AF_INET6, "::1", &sin6_addr), sin6_flowinfo=0, sin6_scope_id=0}, 28) = -1 EAFNOSUPPORT (Address family not supported by protocol)
```

vodik:
And there's no obvious reason why I see it working at 9b39e59accbad9193a6f2bdd55d02c5f0a0e730f but failing on 6aec4e8f2cfdf02498d48d7dc0d1a982845016ec... especially considering how trivial that patch is.

vodik:
Okay, debugging this, looks like the logic for setting `local_ip_is_ipv6` is bypassed for some reason and we never detect it correctly.

vodik:
Actually, looks like `open_connections` can fill the local address information with an IPv4 address even if we're connecting to a remote host.

vodik:
Okay, looks like the fix is simple:

1. We should use the remote socket's family when trying to determine a local sockaddr.
2. Replace the `gethostname(hostname, 64)` with a straight `"localhost'` as a fallback instead.

I need to look into this a little deeper... looks like there are some really weird assumptions in here though. I'll get a patch up later night hopefully.

@rkday how's your branch coming along? If I dive into this, I'd rather not redo the good work you're doing.

wdoekes:
Thanks for looking into this. I added a quick regression test.

rkday:
I've updated https://github.com/SIPp/sipp/pull/148 with my latest code. I'm not sure about replacing `gethostname(hostname, 64)` with `localhost` as a fallback, though - in my /etc/hosts (Fedora 20), localhost is 127.0.0.1 and localhost6 is `::1`, so `localhost` would never resolve to an IPv6 address.

[...]

PR #151, opened 2015-06-20
If the local IP isn't specified, bind as widely as possible

rkday:
@vodik, does this fix https://github.com/SIPp/sipp/issues/150 for you?

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -2521,7 +2521,15 @@ int open_connections()
         struct addrinfo  hints = {AI_PASSIVE, AF_UNSPEC,};
 
         if (!strlen(local_ip)) {
-            local_host = (char *)hostname;
+            if (sendMode == MODE_CLIENT) {
+                if (remote_sockaddr.ss_family == AF_INET) {
+                    local_host = strdup("0.0.0.0");
+                } else {
+                    local_host = strdup("::");
+                }
+            } else {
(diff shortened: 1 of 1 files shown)


src/socket.cpp
```diff
@@ -2521,7 +2521,15 @@ int open_connections()
         struct addrinfo  hints = {AI_PASSIVE, AF_UNSPEC,};
 
         if (!strlen(local_ip)) {
-            local_host = (char *)hostname;
+            if (sendMode == MODE_CLIENT) {
+                if (remote_sockaddr.ss_family == AF_INET) {
+                    local_host = strdup("0.0.0.0");
+                } else {
+                    local_host = strdup("::");
```
vodik: `local_host` can actually be a `const char*`, so you can change this to `local_host = "::";` and drop the casts. Otherwise this will leak memory.

vodik:
Yeah, actually, this is way more correct than what I do than the localhost vs hostname. Merge it!

vodik:
I think `getaddrinfo(NULL, <service>)` will do that same thing without the if statement, but I'll tackle that after your refactoring lands, getting rid of the manual `sin_port`/`sin6_port` magic is good for maintainability.

vodik:
:+1: works perfectly without having to worry about setting an `ai_family` hint.

PR #156, opened 2015-06-23
Fix a few memory leaks at shutdown.

wdoekes:
To run valgrind:

    # make sipp without optimization
    make CFLAGS="-g -O0" CXXFLAGS="-g -O0"

    # run sipp with valgrind
    valgrind --partial-loads-ok=yes --show-leak-kinds=all \
      --leak-check=full \
      sipp SIPP_ARGUMENTS | cat

    # (the extra cat disables curses, which can show distracting leaks)

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -335,24 +335,13 @@ const char *sip_tls_error_string(SSL *ssl, int size)
 
 #endif
 
-char* get_inet_address(struct sockaddr_storage* addr)
+static char* get_inet_address(const struct sockaddr_storage* addr, char* dst, int len)
 {
-    static char* ip_addr = NULL;
-
-    if (!ip_addr) {
-        ip_addr = (char *)malloc(1024*sizeof(char));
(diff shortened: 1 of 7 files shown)


src/socket.cpp
```diff
@@ -335,24 +335,13 @@ const char *sip_tls_error_string(SSL *ssl, int size)
 
 #endif
 
-char* get_inet_address(struct sockaddr_storage* addr)
+static char* get_inet_address(const struct sockaddr_storage* addr, char* dst, int len)
```
vodik:
```c
static char* get_inet_address(struct sockaddr_storage& addr, char* dst, int len)
```
Might be a little nicer.

rkday:
Should we turn those "to run with valgrind" instructions into a Makefile target (make vg_run, make vg_test?).

wdoekes:
> Should we turn those "to run with valgrind" instructions into a Makefile target (make vg_run, make vg_test?).

I added a regression test instead.

PR #157, opened 2015-06-23
More refactoring, cleanups.

vodik:
Putting this stuff up again now so I can get some feedback:

1. Few more internal lists get ported to `std::vector`.
2. Dropped unused macros and functions.
3. No more casting string literals to `char*`.
4. `isWellFormed` is now refactored, unit tested, and deduplicated (was found verbatim in both stat.cpp and scenario.cpp)
5. `setFileName` and friends are also similarly refactored.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -142,7 +142,6 @@ gtest_SOURCES = \
 	gtest/src/gtest-test-part.cc \
 	gtest/src/gtest-typed-test.cc \
 	gtest/src/gtest.cc \
-	gtest/src/gtest_main.cc \
 	gmock/src/gmock-internal-utils.cc \
 	gmock/src/gmock-matchers.cc
 
--- a/include/call.hpp
+++ b/include/call.hpp
@@ -108,7 +108,7 @@ public:
     static int stepDynamicId;   // step of increment for dynamicId
     static int dynamicId;       // a counter for general use, incrementing  by  stepDynamicId starting at startDynamicId  wrapping at maxDynamicId  GLOBALY
 
-private:
+protected:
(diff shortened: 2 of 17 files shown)


vodik:
Okay, so I'm pushing all my work here, even though it overlaps with #158. I will keep this in sync with what we decide there.

vodik:
Quick note with the new tests, they're the way they are to make sure the existing behaviour is consistent, even if I don't think its correct. There are some behaviour differences on bad inputs between `integer_table` and `string_table`:

~~~c++
ASSERT_THAT(string_table_alloc(","), ElementsAre("", ""));
~~~

vs

~~~c++
ASSERT_THAT(integer_table(","), IsEmpty());
~~~

vodik:
Do we have a list somewhere of platforms we consider officially supported? I see two `#ifdef __osf__` branches but something tells me that I doubt that's everything to get this to compile on Tru64 UNIX (which, according to Wikipedia, isn't even an officially supported operating system anymore).

vodik:
Okay, so I hope to leave this as is until someone gets a chance to look over it.

This patch set now completely overhauls SDP parsing (with added tests as well) so that its completely AF-independent. No more manual poking of the `sockaddr_in`/`sockaddr_in6` structures anymore.

Once @rkday's cleanups land, I want to see how much of this code and be unified.

It also kills a lot more manual memory operations, some code duplication, and more cleanups. Its actually a lot less code than the diff stat suggests because I've added quite a few more tests (basically a test for anything I've directly modified).

vodik:
See #149, I'll put this all up together when I've gotten more comfortable with testing everything I change.

PR #158, opened 2015-06-24
Add googlemock support

vodik:
I figured it be useful since we already unit test with googletest. I need this currently for one of my tests on my devel branch:

```c
auto table = integer_table("1,2,3");
ASSERT_THAT(table, testing::ElementsAre(1, 2, 3));
```

But being able to mock classes would be generally useful in expanding the scope of our existing tests.

This also updates the gtest submodule and fixes the annoying repo-is-dirty-because-gtest-is-dirty issue.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -20,3 +20,4 @@
 	#
 	path = gtest
 	url = https://chromium.googlesource.com/external/googletest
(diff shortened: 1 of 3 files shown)


rkday:
Looks good.

PR #161, opened 2015-07-03
Ethernet 802.1Q tagged frames processing.

hpaskalev:
Currently SIPp doesn't handle 802.1Q tagged frames. This patch provides handling of such frames and correct Ethernet type extraction.

### DIFF
--- a/src/prepare_pcap.c
+++ b/src/prepare_pcap.c
@@ -85,6 +85,37 @@ inline uint16_t checksum_carry(int s)
 
 char errbuf[PCAP_ERRBUF_SIZE];
 
+/* get bytes to EtherType block
+*/
+size_t get_ethertype_block_offset(int link, const uint8_t* pktdata)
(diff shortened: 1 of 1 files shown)


vodik:
Hey, so looks good, just one things so we can merge it.

Have a [style guide](https://github.com/SIPp/sipp#contributing), if you could get this code to conform to it, it be awesome (mainly looking at your conditionals). And unless someone else is going to overrule me, I'm also not a fan of the `0x0800 != eth_type` style of comparison. I understand why people code like that, but it reads nicer, imho, the other way.

hpaskalev:
I don't mind pushing another reformat. :)

vodik:
It probably be a good idea to squash all your commits down to a single commit. I'll leave it someone else to review further.

hpaskalev:
Sure, it's a good idea.

PR #167, opened 2015-07-22
Refactor sockaddr_storage and getaddrinfo logic

vodik:
This builds on one of my old devel commits and on @rkday work from #151.

- Kill all the `SOCK_ADDR_SIZE` macros since they're not needed. `getaddrinfo` already tells us the size in its structures.
- Refactor the `getaddrinfo` logic into a single function, `gai_getsockaddr`. It's a verbose block of code that doesn't need to be repeated all over. This also kills the ugly `if (ss_family == AF_INET/AF_INET6)` logic for setting the port.
- Improve the error reporting by appending `gai_strerror` messages.

This also brings improves @rkday's important fix since we can just getaddrinfo assumes a `NULL` for the node argument is equivalent to the 0 ip.

Putting this up for feedback, don't merge yet, its not fully tested :wink:

### DIFF
--- a/include/socket.hpp
+++ b/include/socket.hpp
@@ -144,5 +144,6 @@ void sipp_close_socket(struct sipp_socket *socket);
 #define _RCAST(type, val) ((type)(val))
 #endif
 
+int gai_getsockaddr(struct sockaddr_storage* ss, const char* host, short port, int flags, int family);
 
 #endif /* __SIPP_SOCKET_H__ */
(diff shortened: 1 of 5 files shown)


wdoekes:
What's the status on this? I see it fails to auto-merge.

vodik:
It breaks in one place. I got to investigate it a bit more. I've finally got some breathing room to look into this stuff again :grinning:

vodik:
I've manually merged on the first commit of this patch set onto master, I'm going to abandon the rest of the changes here and try again from scratch.

It introduces a nasty bug that I can't quite figure out where it was introduced. I will start over with smaller, more testable changes

PR #171, opened 2015-07-28
Explicitly bind tcp sockets to source port

vodik:
Quick workaround to get binding working. Also set `SO_REUSEADDR` for good measure.

Not the prettiest code but I want to get this in with minimal rework. I already want to try and change as much as I can to be AF_ independent and this will get cleaned up with that work.

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -2722,6 +2722,25 @@ int open_connections()
         }
         sipp_customize_socket(tcp_multiplex);
 
+        if (user_port) {
+            int sock_opt = 1;
+
+            if (setsockopt(tcp_multiplex->ss_fd, SOL_SOCKET, SO_REUSEADDR, (void *)&sock_opt,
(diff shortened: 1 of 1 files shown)


ramkisuni:
In sipp-master, if we remove the commit 337f5c0, then both -i IPADDR -p PORT is taken into account; with this commit included in the build, if we give -p option it is taken but IPADDR is switched to Primary IP address in the system and not the one we have given with -i option;

ramkisuni:
Using sipp-master, tested UDP,TCP,SCTP and TLS cases with two local changes  in our test environment where sipp was installed in ubuntu 16.04 LTS
1. for TLS refer https://github.com/SIPp/sipp/issues/282 for a local fix (this seems sure a issue)
2. commented out commit 337f5c0 so that -i option and -p option values are considered (it could be my misunderstanding, because the commit was with doubt,I just tried commenting out and found it working for us)
Test results were successful

PR #177, opened 2015-10-08
Do not use std::bind

RichieB2B:
Since include/stat.hpp contains "using namespace std;" calling bind calls std::bind of C++11. That will give errors (on OS X El Capitan) like
```
src/socket.cpp:523:13: error: invalid argument type '__bind<int &, sockaddr *, unsigned long>' to unary expression
        if (!bind(sock, (struct sockaddr *)&ctl_sa, sizeof(struct sockaddr_in))) {
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/socket.cpp:1491:14: error: assigning to 'int' from incompatible type '__bind<int &, sockaddr *, int &>'
    if ((ret = bind(socket->ss_fd, (sockaddr *)saddr, len))) {
             ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
This patch fixes those.

### DIFF
--- a/src/rtpstream.cpp
+++ b/src/rtpstream.cpp
@@ -766,7 +766,7 @@ static int rtpstream_get_localport (int *rtpsocket, int *rtcpsocket)
       (_RCAST(struct sockaddr_in *,&address))->sin_port=
         htons((short)port_number);
     }
-    if (bind(*rtpsocket,(sockaddr *)(void *)&address,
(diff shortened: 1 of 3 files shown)


wdoekes:
No discussion needed there. Thanks!

Issue #173, opened 2015-08-31
sipp.1 man page duplicated


McTrk:
The entire text body of file `sipp.1` is repeated in the file. All text starting from the second occurrence of "Usage: in the middle of the file is identical to most of the first half.
This is caused by help2man using the `--version` option, which `sipp` does not recognize and prints the help printout instead of version info. As a result, help2man sees the same help text twice, instead of help text followed by version info.

Using help2man -v "-v" would fix it.

Suggested correction is in Makefile.am: Add `-v "-v"` as follows:

```Makefile.am
sipp.1: ./sipp $(sipp_SOURCES)
	$(HELP2MAN) --output=$@ -v "-v" --no-info \
		--name='SIP testing tool and traffic generator' \
		./sipp
```

PR #178, opened 2015-10-19
Fixes Issue #173

McTrk:
where man page is duplicated.


rkday:
What does this change do? From reading https://www.gnu.org/software/help2man/, it looks like it jusst passes the -v (version) option - is that right?

McTrk:
Correct, thats's all it does.

rkday:
Ah - having read your notes in https://github.com/SIPp/sipp/issues/173 I now understand why this works. Thanks!

PR #182, opened 2016-01-18
added dynamic files for audio/image/video pcap scenarios.

adubovikov:


wdoekes:
I see how it could be convenient to pass the file as command line argument, but I don't really like how it simply overrides any set value.

I'd rather see something like this:
```
if (strcmp(ptr, "-") == 0) {
    if (!play_pcap_audio) {
        ERROR("play_pcap_audio is supposed to be passed on the command line");
    }
    ptr = play_pcap_audio;
}
tmpAction->setPcapArgs(ptr);
```

Further, `SIPP_OPTION_INT` looks wrong to me. Shouldn't that be SIPP_OPTION_STRING or something?

And, if you could squash the commits, it'd be nice. They seem small and related enough to be combined into a single commit.

Thanks for contributing!

adubovikov:
Hi , if you see, there is the second patch with correct OPTION_STRING, if you want I can recreate it as one.

PR #183, opened 2016-01-18
added dynamic pcap files for audio,image and video rtp scenarios

adubovikov:


adubovikov:
here you are.

include/sipp.hpp
```diff
@@ -260,6 +260,9 @@ extern int                max_rtp_port            _DEFVAL(DEFAULT_MAX_RTP_PORT);
 extern int                rtp_default_payload     _DEFVAL(DEFAULT_RTP_PAYLOAD);
 extern int                rtp_tasks_per_thread    _DEFVAL(DEFAULT_RTP_THREADTASKS);
 extern int                rtp_buffsize            _DEFVAL(65535);
+extern char              * rtp_pcap_audio_file;
+extern char              * rtp_pcap_image_file;
+extern char              * rtp_pcap_video_file;
```
wdoekes: This is in the RTP_STREAM #ifdef. I believe this is specific to for PCAPPLAY.

Please ensure that it's surrounded by the right #ifdefs.

src/scenario.cpp
```diff
@@ -1574,14 +1574,33 @@ void scenario::parseAction(CActions *actions)
                 tmpAction->setIntCmd(type);
 #ifdef PCAPPLAY
             } else if ((ptr = xp_get_value("play_pcap_audio"))) {
+
```
wdoekes: Excess linefeed.

src/scenario.cpp
```diff
@@ -1574,14 +1574,33 @@ void scenario::parseAction(CActions *actions)
...
+                if (strcmp(ptr, "-") == 0) {
+                    if (!rtp_pcap_audio_file) {
+                            ERROR("play_pcap_audio is supposed to be passed on the command line");
```
wdoekes: Excess indent. (Two more below.)

wdoekes:
Very well. A few minor issues as pointed out to fix, and it's ready to go. (Ammend-commit and forcepush on your branch to update this PR.)

I just wonder if we can do a better job at documenting this feature, but I don't have an immediate answer to that. So I won't block it on that.

PR #184, opened 2016-01-18
added dynamic pcap files for audio,image and video rtp scenarios

adubovikov:
IMHO at this time should be ok.


wdoekes:
I wasn't entirely happy with adding additional options. Does #185 fix your issue for you?

PR #185, opened 2016-01-19
Allow play_pcap_audio to take a -key(word) argument.

wdoekes:
You can now do:

    <exec play_pcap_audio="[file1]"/>
    <exec play_pcap_image="[file1]"/>
    <exec play_pcap_video="[file1]"/>

Pass `-key file1 path/to/file1.pcap` on the command line.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,14 @@
 Features added since 3.5.0
 ==========================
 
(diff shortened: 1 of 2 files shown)


adubovikov:
yep, here is also a good way! lets do it1

PR #189, opened 2016-02-08
RFC: Enable basic Travis-CI support

lemenkov:
This enables basic Travis-CI support. It builds with both GCC and CLang on Ubuntu 12.04 and runs make check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,27 @@
+language: c
+compiler:
+        - gcc
+        - clang
+
+notifications:
+        email: false
+
+python:
+        - "2.7"
+
+sudo: required
+
+before_install:
+        - sudo apt-get -qq update
+        - sudo apt-get install -y libpcap-dev libsctp-dev libncurses5-dev libssl-dev libgsl0-dev
+
+before_script:
+        - git submodule update --init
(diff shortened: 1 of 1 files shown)


lemenkov:
Here is a test build

https://travis-ci.org/lemenkov/sipp/builds/107785842

wdoekes:
Sounds like a good idea to me :+1: 

Bonus points if you can automatically run `./regress/runtests` as well. I skipped a few tests just now that fail for whatever reason, so it should return 0. (You may need add another dependency on perl.)

wdoekes:
... and a status image in the README: https://api.travis-ci.org/.../sipp.svg

lemenkov:
So, what others think? Shall we just merge it and continue further improving it?

wdoekes:
Let's do that :)

lemenkov:
@rkday could you please enable Travis-CI integration? Only one with admin powers can do this. :)

wdoekes:
Never mind @rkday. I've got it.

wdoekes:
Nice. Looks like it works. Even with the regression tests without any hassle.

lemenkov:
@wdoekes Cool! Now we should add more testing! :)

PR #204, opened 2016-03-22
Refactor sockaddr code for clarity

vodik:
Add two helpers, one wrapping the copypasta `getaddrinfo` logic, and one for wrapping the sockaddr port changing logic.

Seems to work under testing, makes the sockaddr code a lot easier to understand. Hopefully more refactoring will fall out as a result.


wdoekes:
Looks good to me. And my test runs still succeeded with it.

vodik:
Oh shit, looks like I don't have my new editor setup correctly there. Crap. Thanks!

PR #209, opened 2016-03-24
play_dtmf: Add filler packets to warm up the RTP stream.

### LINKED ISSUE #208: first two play_dtmf tones appear to not respect length parameters
When testing https://github.com/ossobv/sipp-scenarios/blob/master/dtmf/playdtmf.xml it appears that the first two tones in the play_dtmf block don't get a fair wait time.

When setting tones to 12345, it sounds like 12-3-4-5.

wdoekes:
(closes #208)

The gist of this commit is in the "needs_filler" code path: it adds 400ms of no-op stream before every DTMF so the stream gets warmed up.

In my case, if I call my SPA941 (through Asterisk) the first two audible DTMF tones get concatenated if I disable the needs_filler code.

### DIFF
--- a/src/prepare_pcap.c
+++ b/src/prepare_pcap.c
@@ -37,6 +37,11 @@
 #include "endianshim.h"
 #include "prepare_pcap.h"
 
+/* Helpful RFCs for DTMF generation.
+ * https://tools.ietf.org/html/rfc4733
+ * https://tools.ietf.org/html/rfc3550
+ */
+
 /* We define our own structures for Ethernet Header and IPv6 Header as they are not available on CYGWIN.
  * We only need the fields, which are necessary to determine the type of the next header.
  * we could also define our own structures for UDP and IPv4. We currently use the structures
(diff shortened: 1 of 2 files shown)


src/prepare_pcap.c
```diff
@@ -274,13 +279,14 @@ int prepare_pkts(const char* file, pcap_pkts* pkts)
...
         }
+        /* BUG: inefficient */
         pkts->pkts = (pcap_pkt *)realloc(pkts->pkts, sizeof(*(pkts->pkts)) * (n_pkts + 1));
         if (!pkts->pkts)
             ERROR("Can't re-allocate memory for pcap pkt");
         pkt_index = pkts->pkts + n_pkts;
         pkt_index->pktlen = pktlen;
         pkt_index->ts = pkthdr->ts;
-        pkt_index->data = (unsigned char *) malloc(pktlen);
+        pkt_index->data = (unsigned char *) malloc(pktlen); /* BUG: inefficient */
```
vodik: Yeah, I looked at refactoring this when I merged the code in initially. Problem is that you don't have enough information for the packet generation until the last minute, unlike with pcap_play - at least without some deeper work.

src/prepare_pcap.c
```diff
@@ -349,53 +355,227 @@ struct dtmfpacket {
...
+    udp->uh_dport = 0;
 #else
-    dtmfpacket->udp.len = htons(pktlen);
-    dtmfpacket->udp.check = 0;
-    dtmfpacket->udp.source = 0;
-    dtmfpacket->udp.dest = 0;
+    udp->len = htons(pktlen);
+    udp->check = 0;
+    udp->source = 0;
+    udp->dest = 0;
```
vodik: This is another thing (out of scope) if we compile with `DEFAULT_SOURCE`, you get BSD style definitions. We don't need this if/else.

PR #212, opened 2016-03-28
Move new sockaddr helpers out of the USE_OPENSSL section

rkday:
I noticed that master didn't compile - this was because I didn't have OpenSSL and a couple of the functions added recently were in an `#ifdef USE_OPENSSL` block.

This moves them out of that block and makes that block more obvious (removing some newlines in it and commenting the end).


wdoekes:
Good. These checks should cover that for the future: https://github.com/SIPp/sipp/pull/213

PR #214, opened 2016-03-29
Refactor socket.cpp into a SIPpSocket class

rkday:
I've done a chunk of refactoring to turn socket.cpp, with its sipp_socket struct and bunch of methods that operate on it, into a SIPpSocket class. This has as many fields/methods private as I could manage - hopefully this makes the interface a bit more comprehensible.

I'd like to refactor further (maybe having UDP/TCP/etc. subclasses rather than an ss_transport field?) but I think this is a good checkpoint.


wdoekes:
The changes look sane to me, apart from some indentation that could be improved here and there.

I was also looking at the socketowner class, wondering whether we could pass a const socket there in a few cases, but that's for another time.

Note that I did update the indentation checks through travis, so you'll have to rebase against master and fix new issues before merging.

Note to self: update travis-matrix to also check with/without HAVE_EPOLL.

src/socket.cpp
```diff
@@ -2931,214 +2931,214 @@ void SIPpSocket::pollset_process(int wait)
     for (int event_idx = 0; event_idx < rs; event_idx++) {
         int poll_idx = (int)epollevents[event_idx].data.u32;
 #else
-    for (size_t poll_idx = 0; rs > 0 && poll_idx < pollnfds; poll_idx++) {
+        for (size_t poll_idx = 0; rs > 0 && poll_idx < pollnfds; poll_idx++) {
```
wdoekes: This indent is wrong; confused by the HAVE_EPOLL ifdef.

src/socket.cpp
```diff
@@ -95,14 +95,14 @@ int gai_getsockaddr(struct sockaddr_storage* ss, const char* host,
...
     switch (ss->ss_family) {
-    case AF_INET:
-        _RCAST(struct sockaddr_in*, ss)->sin_port = htons(port);
-        break;
-    case AF_INET6:
-        _RCAST(struct sockaddr_in6*, ss)->sin6_port = htons(port);
-        break;
-    default:
-        ERROR("Unsupported family type");
+        case AF_INET:
```
wdoekes: I'm not fond of indented case myself.
And I prefer python-rules for indenting: 4 spaces unless it conflicts with readability: after `if (` you take 8 spaces so the if-condition stands out from the if-body.

Did you use some auto-indent on this file? It didn't work out so well at the end ;)

vodik:
Looks good to me too.

lemenkov:
Cool! Let's merge it!

wdoekes:
Rebased and force-pushed.

wdoekes:
And another force-push, removing the mess that 672664b1 made.

PR #216, opened 2016-03-30
Refactor SDP parsing (with more unit testing!)

vodik:
Refactor the bizarrely over complicated sockaddr poking magic of the SDP
parser into two functions that reuse the new gai_getsockaddr helpers.

Bonus mockcall object and unit tests for the new parsing logic. Very
rudimentary, but should do the trick and lay some groundwork.

### DIFF
--- a/src/call.cpp
+++ b/src/call.cpp
@@ -123,204 +123,52 @@ unsigned int call::wake()
 }
 
 #ifdef PCAPPLAY
-static char* find_sdp_eol(char* line)
+static std::string find_in_sdp(std::string const &pattern, std::string const &msg)
 {
-    char* end = &line[strcspn(line, "\r\n")];
-    return end[0] == '\0' ? NULL : end;
+    std::string::size_type begin, end;
+
+    begin = msg.find(pattern);
+    if (begin == std::string::npos) {
+        return "";
+    }
+
+    begin += pattern.size();
+    end = msg.find_first_of(" \r\n", begin);
+    if (end == std::string::npos || begin == end) {
+        return "";
+    }
+
+    return msg.substr(begin, end - begin);
 }
 
-/******* Media information management *************************/
-/*
- * Look for "c=IN IP4 " pattern in the message and extract the following value
- * which should be IP address
- */
-uint32_t get_remote_ip_media(char *msg)
+void call::get_remote_media_addr(std::string const &msg)
 {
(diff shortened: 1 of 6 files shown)


src/call.cpp
```diff
@@ -4228,3 +4077,97 @@ void *send_wrapper(void *arg)
...
+    reference.sin6_family = AF_INET6;
+    reference.sin6_port = htons(12345);
+    inet_pton(AF_INET6, "::1", &reference.sin6_addr);
+
+    mockcall call(true);
+    call.parse_media_addr(test_sdp_v6);
+    ASSERT_EQ(call.has_media(), true);
+    ASSERT_EQ(reference, call.get_audio_addr<struct sockaddr_in6>());
+}
+#endif
```
wdoekes: A `/* PCAP_PLAY */` after this endif helps readability.

src/call.cpp
```diff
@@ -123,201 +123,50 @@ unsigned int call::wake()
 }
 
 #ifdef PCAPPLAY
-static char* find_sdp_eol(char* line)
+static std::string parse_sdp_msg(std::string const &pattern, std::string const &msg)
```
wdoekes: A bit odd naming perhaps? Should it be called `find_in_sdp` or `get_sdp_line` (or similar) instead?

src/call.cpp
```diff
@@ -123,201 +123,50 @@ unsigned int call::wake()
...
-
-    if (pattype == PAT_AUDIO) {
-        pattern = "m=audio ";
-    } else if (pattype == PAT_IMAGE) {
-        pattern = "m=image ";
-    } else if (pattype == PAT_VIDEO) {
-        pattern = "m=video ";
-    } else {
-        ERROR("Internal error: Undefined media pattern %d\n", 3);
+    std::string port, host = parse_sdp_msg(media_ip_is_ipv6 ? "c=IN IP6 " : "c=IN IP4 ", msg);
```
wdoekes: Declaring the port here is confusing, looks almost like parse_sdp_msg returns a tuple (with my python eyes).

vodik:
Weird, its complaining of tabs in src/call.cpp, but I can't find them and `validate-src.sh` locally tells me I'm clean...

wdoekes:
Nice work :+1:

vodik:
Alright, merging it.

PR #220, opened 2016-04-12
Use DEFAULT_SOURCE for BSD style packet structs

vodik:
Reduces a lot of the needless duplication that is caused by difference in structure implementations.

Note that when targeting gnu11, DEFAULT_SOURCE is actually implied, but I'm adding it just in case.


wdoekes:
Nice :)

Issue #224, opened 2016-04-21
Potential use-after-free bugs in SIPpSocket


vodik:
1. `SIPpSocket::invalidate` may trigger `SIPpSocket::abort`, which does `delete this`. However this isn't signalled to the calling context, leading to use after free bugs. So `SIPpSocket::read_error` calls `SIPpSocket::invalidate` and then may continue to use class methods (`SIPpSocket::close_calls`).
2. `SIPpSocket::reconnect` calls `SIPpSocket::abort` if `ss_invalid` is set. It may then proceeds to access deleted internals like `ss_ipv6`.

wdoekes:
Fixed, right?

PR #225, opened 2016-04-21
Fix defects

vodik:
Fix some new defects that popped up while reworking the pcap_play and SIPpSocket stuff.

### DIFF
--- a/src/call.cpp
+++ b/src/call.cpp
@@ -1260,6 +1260,12 @@ bool call::executeMessage(message *curmsg)
         }
 
         msg_snd = send_scene(msg_index, &send_status, &msgLen);
+        if (!msg_snd) {
+            /* This will hit connect_if_needed, and if it fails, the
+               entire call is deleted... */
+            ERROR("Call failed, cannot continue safely...");
+        }
+
         if(send_status < 0 && errno == EWOULDBLOCK) {
             if (incr_cseq) --cseq;
             /* Have we set the timeout yet? */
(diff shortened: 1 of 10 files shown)


src/call.cpp
```diff
@@ -1260,6 +1260,12 @@ bool call::executeMessage(message *curmsg)
         }
 
         msg_snd = send_scene(msg_index, &send_status, &msgLen);
+        if (!msg_snd) {
+            /* This will hit connect_if_needed, and if it fails, the
+               entire call is deleted... */
+            ERROR("Call failed, cannot continue safely...");
```
vodik: This is actually also related to #224.

src/scenario.cpp
```diff
@@ -1857,18 +1860,17 @@ int createStringTable(char * inputString, char *** stringList, int *sizeOfList)
     *stringList = NULL;
     *sizeOfList = 0;
 
+    /* FIXME: temporary workaround: needs rewrite */
+    char *p = const_cast<char*>(strchr(inputString, ','));
```
vodik: I have a rewrite of this method (porting it to std::vector) sitting around, I'll try to get it back in. In the meanwhile, this is at least safe - if very hacky.

src/scenario.cpp
```diff
@@ -236,7 +236,7 @@ double get_double(const char *ptr, const char *what)
 /* Return static buffer to xml value, as with xp_get_value().
  * If the value is enclosed in [brackets], it is assumed to be
  * a command-line supplied keyword value (-key). */
```
wdoekes: The comment is wrong. Static buffer?

wdoekes:
LGTM, apart from the stale static-buffer comment.

src/scenario.cpp
vodik: Right, didn't notice that.

PR #227, opened 2016-04-28
Fix incorrect packet length

### LINKED ISSUE #223: RTP play with sipp 3.6 is broken or please let me know why it works with 3.3.990
I have spent some time trying to send a simple .pcap audio stream using sipp version 3.6 without success, my server is running uas.xml file unmodified, the command for server is

sipp -sf /home/myusernamehere/Devel/sipp-scenarios/uas.xml -mi 192.168.3.7

the client is using the uac_pcap.xml file unmodified too, the one that sends the g711a.pcap and dtmf_2833_1.pcap.

The client command is

/sipp -sf /home/myusernamehere/Devel/SIP/sipp-scenarios/uac_pcap.xml 192.168.3.7 -mi 192.168.3.5 -m 1

Using same commands and scenarios but with version 3.3.99 it works. 

The way I am determining it doesn't work is capturing outgoing traffic with tcpdump and reading file in wioreshark, the version 3.6 seems to send RTP packets with all zeros in UDP/RTP/Payload and so length is 36780, also I notice that for some reason the packet numbering in wireshark and the counted by tcpdump as captured is ridiculous big, for the example above it should be 251 or so, and with version 3.6 the count is 5663, a further inspection shows that there are the same number of RTP packets (bad as explained before) but numbering on wireshark makes jumps in sequence to get to 5k.

Also, one more thing that might be wrong is that when running client with version 3.6 is claims about some packet bigger that some define, the error is

2016-04-19      18:24:42.274355 1461104682.274355: Packet size is too big! Recompile with bigger PCAP_MAXPACKET in prepare_pcap.h.

I increase the size in definition (don't blame me if this is the reason but rather let me know how to fix the error) in

#define PCAP_MAXPACKET 1500

to some bigger number and it allow me to run.

Version 3.3.99 don't suffer from this at all.

Don't know how to upload the captures, as .zip the page won't accept it.

vodik:
My bad, this looks like something caused by 4ae32a09c8fcd5e0b88b727e53cf96bd5e94433a (*why* is there a +4 hidden in there? that's not correct... unless its to compensate for structure padding).

Which leads to some questions:

- Why do we use the system's network structures for everything else but ip4?
- Did this code work on BSD/Cygwin (maybe we should add an OSX travis build)?

Now while I'm putting this up for review, I don't know if its ready to be merged. Its started causing some of our internal SIPp based tests to start to fail, claiming that DTMF packets are getting detected where we previous weren't. But now, and especially after looking at this code in depth, I wouldn't be surprised if it never worked. The packet payload offset calculation on ipv4 looks suspiciously broken.

---

Looks like the ipv4 packet length code is completely wrong.

The original code seems to be doing special logic on OSX, Cygwin and
FreeBSD, then again on HP UX and then again on Linux (and everything
else) because of, what I can only guess, must be differences in
structure padding. Mark the iphdr struct as packed for good measure.

There is no reason for the + 4, it shouldn't have slipped in.

Probably should use the system provided structures, like we're already
doing for ipv6. But one thing at a time.

Fixes #223.

### DIFF
--- a/include/prepare_pcap.h
+++ b/include/prepare_pcap.h
@@ -45,7 +45,7 @@ struct iphdr {
     u_int32_t saddr;
     u_int32_t daddr;
     /*The options start here. */
-};
+} __attribute__((packed));
 
 #endif
 
--- a/src/prepare_pcap.c
+++ b/src/prepare_pcap.c
@@ -252,7 +252,6 @@ int prepare_pkts(const char* file, pcap_pkts* pkts)
         iphdr = (struct iphdr*)((char*)ethhdr + sizeof(*ethhdr));
         if (iphdr && iphdr->version == 6) {
             /* ipv6 */
-            pktlen = (uint64_t)pkthdr->len - sizeof(*ethhdr) - sizeof(*ip6hdr);
             ip6hdr = (ipv6_hdr*)(void*)iphdr;
             if (ip6hdr->nxt_header != IPPROTO_UDP) {
                 fprintf(stderr, "prepare_pcap.c: Ignoring non UDP packet!\n");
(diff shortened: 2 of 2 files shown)


vodik:
Ah, I know where this is broken now. I've tweaked the ipv4 code to look like the ipv6 code... but the ipv6 code is broken. `sizeof(*ethhdr)` is 2... which is wrong. The `ethhdr` struct is not the complete structure, you need to poke at `ether_type_offset`...

So yeah, I should be able to get this fixed up... and then look at getting some tests for it.

vodik:
Okay, I'm overthinking it. Just always forward the length the udp packet itself reports, in all cases. Should fix both ipv4 and ipv6.

wdoekes:
:+1: 

Yup. Reports the same numbers as with 3.5, so LGTM.

vodik:
Yeah, its probably the only correct way.

The code in prepare_pcap.c assumes that the udp packet size will reliably be determined by the size of the packet minus the size of the datalink and network layer frames, but that's not guaranteed, since the code ignores that ethernet packets smaller than 60 bytes are padded out. And we do have to worry about this rule because most rtpevent packets are that small.

PR #229, opened 2016-04-29
Improve Travis CI jobs and add OSX build

vodik:
I personally don't run anything other than Linux. Travis CI does offer doing OSX builds, so adding it to the continuous integration build should give us some more confidence changing on non-Linux platforms.

Also switched over to the new container style stuff (https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/), for much faster builds on Linux (4min instead of about 12). Took some time and try to clean things up a little bit too (for example, checking out submodules is redundant).

Only catch is the regress scripts don't run on OSX.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,13 +51,19 @@ matrix:
         - CC=gcc-4.6
         - CXX=g++-4.6
         - CONFOPTS="--with-gsl --with-openssl --with-pcap --with-rtpstream --with-sctp"
+    - os: osx
+      compiler: clang
+      env:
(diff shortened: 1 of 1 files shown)


vodik:
Also looks like regress scripts can get a little stuck under docker... Need to investigate whats going on.

vodik:
I'm just going to merge this so I can start and get on with other refactoring work.

wdoekes:
Yes, sure thing. I thought you were stalled on getting the scripts unstuck. Merging this is fine with me.

vodik:
Oh, yeah I see. No I have no idea whats wrong - I'll have to catch it in action again.

PR #232, opened 2016-05-31
Remove old portability shims

vodik:
So, putting this up for review, but still need to manually test it (don't know if this won't break pcap_play, etc at the moment).

There are a lot of portability shims in the networking code. We shouldn't need to redefine `struct ip` or `struct ip6_hdr` on any platform I know about anyways. Cygwin seems to have full ipv6 support these days, for example.

### DIFF
--- a/include/prepare_pcap.h
+++ b/include/prepare_pcap.h
@@ -22,33 +22,6 @@
 #include <netinet/udp.h>
 #include <time.h>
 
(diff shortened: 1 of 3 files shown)


wdoekes:
Looks sane to me.

lemenkov:
I like PRs which remove stuff instead of adding it!

vodik:
So no regressions on our internal sipp stuff, so I think I'll merge it and deal with any fallout as it gets reported :grinning:

Issue #243, opened 2016-08-05
Error on TCP connection, remote peer probably closed the socket: SSL_read returned SSL_ERROR_WANT_READ.


sgel83:
SIPp v3.5.1-TLS-SCTP-PCAP-RTPSTREAM built Aug  4 2016, 16:03:38.

Using TLS for a UAC scenario sometimes under high load errors out with the following error.

It seems that currently the software is not handling the SSL_ERROR_WANT_READ as OpenSSL recommends.  It should upon receiving such a response from the remote party try the SSL_READ again, instead of causing a failure.

https://www.openssl.org/docs/manmaster/ssl/SSL_get_error.html
http://stackoverflow.com/questions/3952104/how-to-handle-openssl-ssl-error-want-read-want-write-on-non-blocking-sockets

- Stephen

PR #244, opened 2016-08-14
Fixes for TLS/OpenSSL calls

### LINKED ISSUE #241: High volume TLS causes crash
Hello,

I am trying to do some high load testing in TLS.  It crashes with the following error on the UAC side.

sipp: src/socket.cpp:1895: int SIPpSocket::read_error(int): Assertion `(*__errno_location ()) != 11' failed.
                                                                                                            Aborted

------------------------------ Scenario Screen -------- [1-9]: Change Screen --
  Call-rate(length)   Port   Total-time  Total-calls  Remote-host
5040.0(0 ms)/1.000s   5060      24.04 s        46975  53.53.53.10:5061(TLS)

  5039 new calls during 1.003 s period   0 ms scheduler resolution
  5618 calls (limit 1000000)             Peak was 5743 calls, after 23 s
  1 Running, 46974 Paused, 1098 Woken up
  14022 dead call msg (discarded)        0 out-of-call msg (discarded)
  4 open sockets

                                 Messages  Retrans   Timeout   Unexpected-Msg
      INVITE ---------->         46975     0         0
         100 <----------         22188     0         0         19172
         180 <----------         22166     0         0         0
         183 <----------         0         0         0         0
         200 <----------  E-RTD1 22187     0         0         0
         ACK ---------->         22187     0
       Pause [      0ms]         22186                         0
         BYE ---------->         22185     0         0
         200 <----------         22185     0         0         0

------ [+|-|*|/]: Adjust rate ---- [q]: Soft exit ---- [p]: Pause traffic -----

Last Error: Aborting call on unexpected message for Call-Id '38933-14535...
sipp: src/socket.cpp:1895: int SIPpSocket::read_error(int): Assertion `(*__errno_location ()) != 11' failed.
                                                                                                            Aborted

It seems according to the comments in the code that this shouldn't ever happen.

    /* We have only non-blocking reads, so this should not occur. */
    if (ret < 0) {
        assert(errno != EAGAIN);
    }

rkday:
Fixes #241 and #243.

For #241, the issue was that the OpenSSL functions don't set errno - so checking errno after calling them didn't give valid results.

For #243, we now just return 1 from read_error when we hit SSL_ERROR_WANT_READ - this should avoid the code in read_error which closes the socket and the code at https://github.com/SIPp/sipp/blob/04c38740cbc6476050ca37754f7fdd272e5615bc/src/socket.cpp#L3022 which removes it from the list of file descriptors.

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -1877,7 +1877,14 @@ int SIPpSocket::read_error(int ret)
     const char *errstring = strerror(errno);
 #ifdef USE_OPENSSL
     if (ss_transport == T_TLS) {
+        int err = SSL_get_error(ss_ssl, ret);
         errstring = sip_tls_error_string(ss_ssl, ret);
+        if (err == SSL_ERROR_WANT_READ) {
+            /* This is benign - we just need to wait for the socket to be
+             * readable again, which will happen naturally as part of the
+             * poll/epoll loop. */
+            return 1;
+        }
     }
 #endif
 
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -1890,8 +1890,10 @@ int SIPpSocket::read_error(int ret)
     }
 #endif
 
-    /* We have only non-blocking reads, so this should not occur. */
-    if (ret < 0) {
+    /* We have only non-blocking reads, so this should not occur. The OpenSSL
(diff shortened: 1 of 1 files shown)


lemenkov:
LGTM.

PR #245, opened 2016-08-20
Fix Solaris build

sjthomason:
Define __SUNOS on Solaris

### DIFF
--- a/include/endianshim.h
+++ b/include/endianshim.h
@@ -16,7 +16,10 @@
 /* Darwin does something else. */
 #include <libkern/OSByteOrder.h>
 #endif
-
+#if defined(__SUNOS)
+/* Solaris and derivatives */
+#include <sys/byteorder.h>
+#endif
 
 #if defined(__DARWIN)
 #define le16toh(x) OSSwapLittleToHostInt16(x)
@@ -26,6 +29,13 @@
 #define le16toh(x) ((uint16_t)( \
     (((uint16_t)(x)) << 8) | \
     (((uint16_t)(x)) >> 8)))
+
+#elif defined(__SUNOS)
+#ifdef _BIG_ENDIAN
+#   define le16toh(x) BSWAP_16(x)
+#else
+#   define le16toh(x) (x)
+#endif
 
 #elif !defined(HAVE_DECL_LE16TOH) || HAVE_DECL_LE16TOH == 0
 /* le16toh() is missing in glibc before 2.9 */
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -1433,7 +1433,7 @@ SIPpSocket* SIPpSocket::accept() {
             // Keep the original socket fd.
             newfd = fd;
         } else {
-            close(fd);
+            ::close(fd);
         }
         fd = newfd;
     }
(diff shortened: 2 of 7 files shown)


sjthomason:
Initially I noticed that __SUNOS wasn't being defined on Solaris by autoconf but after digging a little deeper I found some additional build issues.  This PR at least allows a build to complete on Solaris but I have a few questions regarding the projects stance on a few items:
- Can u_intX_t be replaced with standard uintX_t or would a typedef (or CPP define) be preferable on platforms that require it?
- Is is preferable to relay on autoconf to define __SUNOS or the more standard
```
#if (defined(__SVR4) && defined(__sun))
```

wdoekes:
Thanks for the PR!

> Can u_intX_t be replaced with standard uintX_t or would a typedef (or CPP define) be preferable on platforms that require it?

Yes. It'd be better if they were replaced by the standardized types. No one has taken the time to fix this though. Any patches are welcome.

> Is is preferable to relay on autoconf to define __SUNOS or the more standard [...]

Standards are better :)

That being said, I believe there was another fix for the missing le16toh():
https://github.com/SIPp/sipp/commits/fix/3.5/add-missing-le16toh-on-old-glibc-211
Does that work for you?

include/endianshim.h
```diff
@@ -13,6 +13,15 @@
...
 #endif
 
+#ifdef __SUNOS
+#include <sys/byteorder.h>
+#ifdef _BIG_ENDIAN
+#define le16toh(x) BSWAP_16(x)
+#else
+#define le16toh(x) (x)
+#endif
+#endif
```
wdoekes: I merged the endian changes just now. Do those fix this bit for you?

src/socket.cpp
```diff
@@ -1685,7 +1685,11 @@ void SIPpSocket::sipp_sctp_peer_params()
         memset(&peerparam, 0, sizeof(peerparam));
 
         sockaddr* addresses;
+#ifdef __SUNOS
+        int addresscount = sctp_getpaddrs(ss_fd, 0, (void **) &addresses);
+#else
         int addresscount = sctp_getpaddrs(ss_fd, 0, &addresses);
```
wdoekes: How about casting to `(struct sockaddr**)` instead. Then you can drop the `#ifdef`.

sjthomason: Solaris uses a different prototype unfortunately:
int sctp_getpaddrs(int sock, sctp_assoc_t id, void **addrs)

Without the cast to the cast to ```(void **)``` g++ throws an error:
```error: invalid conversion from ‘sockaddr**’ to ‘void**’```

File: include/endianshim.h (hunk 1 above)
Comment (sjthomason) [inline include/endianshim.h]: Unfortunately Solaris needs a different header.  I updated the PR with your recent changes and used the builtin BSWAP_16 macro on this platform.

Thanks!

**Comment by wdoekes** [comment]
Thanks! I cherrypicked/adapted the changes so they could be merge upwards from 3.5. Could not test, but I think it should works.

**Comment by sjthomason** [comment]
Hi @wdoekes,  it appears a necessary change didn't make it to master:
```
diff --git a/src/send_packets.c b/src/send_packets.c
index 12838fb..5610e3f 100644
--- a/src/send_packets.c
+++ b/src/send_packets.c
@@ -39,9 +39,9 @@
 #include <stdlib.h>
 #include <stdbool.h>
 #include <stdint.h>
-#include <netinet/udp.h>
 #include <netinet/in.h>
 #include <netinet/ip6.h>
+#include <netinet/udp.h>
 #include <errno.h>
 #include <string.h>
 #include <fcntl.h>
```

The include order is significant on some platforms:
```/usr/include/netinet/udp.h:26:2: error: unknown type name ‘in_port_t’```

wdoekes:
Drat! Sorry about that =)

Fixed now.

PR #251, opened 2016-09-02
Fix bind/connect/sendto issue because of bad addrlen.

### LINKED ISSUE #246: Unable to bind audio RTP socket
After commit 4aa361d5aefc080d32d96134c9d7493a1eeee564, OS X builds fail to bind with the error:
Unable to bind audio RTP socket (IP=X.X.X.X, port=6100), errno = 22 (Invalid argument).

I did a git bisect and found that the previous commit works correctly.

wdoekes:
In 4aa361d, the SOCK_ADDR_SIZE macro was removed and in bind(),
connect() and sendto() calls, it was replaced with sizeof(struct
sockaddr_storage). Unfortunately some systems (OSX) refuse to work
with an inexact (too large) addrlen.

Thanks to @sjthomason for reporting and bisecting the cause.
Closes #246.

### DIFF
--- a/include/socket.hpp
+++ b/include/socket.hpp
@@ -35,6 +35,32 @@ ssl_init_status FI_init_ssl_context (void);
 
 #endif // USE_OPENSSL
 
+/**
+ * On some systems you must pass the exact sockaddr struct size to
+ * connect/bind/sendto calls. Passing a length that is too large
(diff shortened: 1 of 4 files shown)


pbertera:
Worked for me in OSX 10.10.5.

Thanks,

PR #253, opened 2016-09-21
Added a new action: loadfile.

KalitaAlexey:
This kind of action supports the following attributes:
* assign_to - variable to save file content in
* file_path - file path to load content from

It allows to communicate between SIPp instances


KalitaAlexey:
I did something wrong. It has wrong commit.

PR #254, opened 2016-09-21
Added a new attribute `sync_cmd` on a `exec` action.

KalitaAlexey:
It allows to execute some command and to wait till the command end.

### DIFF
--- a/include/actions.hpp
+++ b/include/actions.hpp
@@ -57,6 +57,7 @@ public:
         E_AT_LOG_WARNING,
         E_AT_LOG_ERROR,
         E_AT_EXECUTE_CMD,
+        E_AT_EXECUTE_SYNC_CMD,
         E_AT_EXEC_INTCMD,
         E_AT_VAR_ADD,
         E_AT_VAR_SUBTRACT,
--- a/src/call.cpp
+++ b/src/call.cpp
@@ -3583,7 +3583,8 @@ call::T_ActionResult call::executeAction(char * msg, message *curmsg)
         } else if (currentAction->getActionType() == CAction::E_AT_LOG_ERROR) {
             char* x = createSendingMessage(currentAction->getMessage(), -2 /* do not add crlf*/);
             ERROR("%s", x);
-        } else if (currentAction->getActionType() == CAction::E_AT_EXECUTE_CMD) {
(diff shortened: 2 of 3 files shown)


wdoekes:
I could see a use for this, but:
- I'd rather see a `<exec wait="true"/>` property on it so we don't add lots of new commands.
- If we're going to wait for the process, I don't think there is a need to double fork, like we do now. You should instead skip the second fork.
- I have no idea what this does to multiple simultaneous calls. Have you tested this?

KalitaAlexey:
@wdoekes You are right about *wait* attribute. This is more convenient.
Could you, please, describe why two forks?
Bad things happen with multiple simultaneous calls: it is going to wait for the command end.
I don't know whether it serious or not.

wdoekes:
> Could you, please, describe why two forks?

The double fork is to daemonize the subprocess. That way SIPp can clean up it's file descriptors and continue running. (You'll find plenty of resources on the internet about this subject.)

> Bad things happen with multiple simultaneous calls: it is going to wait for the command end.
> I don't know whether it serious or not.

If you're going to wait for the command to end, you'd need only one fork and waitpid on that one.

But, I don't find it acceptable if this feature stalls the entire SIPp operation.

You'd need to find a way to delay the call where the command has spawned while resuming SIPp operations. This probably requires:
- maintaning a list of forked processed to reap (at most one per call?)
- not resuming the call until the subprocess has ended and is reaped (See: `man 2 waitpid` and `WNOHANG` and/or select calls on the filedescriptor to wait for activity.)

KalitaAlexey:
@wdoekes Yes, I will do it. I think that should be closed. I will reopen when an update will be ready.

wdoekes:
Very well. Closing for now!

lamplet:
I found that sync_cmd will pause all calls. I want a sync_cmd that will only pause the current call. The scenario is:
1. some sip flow.
2. execute a shell script to send a http request and receive the response.
3. wait the script complete, and continue the sip flow. only this call will wait, other calls are still running.

Maybe we can do it another way:
<action command="somescript.sh taskid"/>
<pause />
and then send a udp packet to wakeup the task by extending the 'Remote control' function.

KalitaAlexey:
Hi @lamplet. I have done this in my fork. I have no time to prepare PR to this repo.

lamplet:
@KalitaAlexey Thank you!

PR #255, opened 2016-09-21
Added a new action: loadfile.

KalitaAlexey:
This kind of action supports the following attributes:
* assign_to - variable to save file content in
* file_path - file path to load content from

It allows to communicate between SIPp instances.

### DIFF
--- a/include/actions.hpp
+++ b/include/actions.hpp
@@ -48,6 +48,7 @@ public:
         E_AT_ASSIGN_FROM_STRING,
         E_AT_ASSIGN_FROM_INDEX,
         E_AT_ASSIGN_FROM_GETTIMEOFDAY,
+        E_AT_LOADFILE,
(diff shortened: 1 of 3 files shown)


wdoekes:
There is already a way to communicate between SIPp instances: `-3pcc`

I'm sorry. I don't think file-based message passing adds anything useful here.

KalitaAlexey:
@wdoekes Could I do the following scenario with -3pcc?
There are two SIPp instances.
There are two SIP devices.
The first device (later D1) is connected to the first SIPp instance (later S1).
The second device(later D2) is connected to the second SIPp instance (later S2).
D1 sends an INVITE request to the D2 through S1.
S1 sends this request to the S2.
S2 sends this request to the D2.
D2 sends response to the S2.
S2 sends response to the S1.
I didn't find any way to do that.

Issue #132, opened 2015-05-14
It'd be nice if [last_Header] without colon meant only the header value.


wdoekes:
Right now `[last_From:]` and `[last_From]` are synonymous: they mean the entire header, including the header name.

Quite often, however, you need to put the from-header value in the To header, and vice versa.

This could be simplified if `[last_From]` meant only the From-header value.

For that change to work, we need a couple of things:
- fix the example scenario's to use colon everywhere
- add the appropriate code to make the distinction
- add an option somewhere to get old-style behaviour (or an option to get new-style?)

Any ideas on that last bit?

I was thinking we could perhaps options to the scenario XML root to accomodate that.

rkday:
Might be a bit of an obscure distinction, and break existing scripts - could we introduce a new `[last_From_value]` instead? That gets the same benefit, and is a bit more obvious at first glance.

vodik:
This is something I've tossed around a bit, I think I'd rather see a dot notation: `[last_From.value]`. This opens up the notion of "direction" (e.g. `[last_From.user]`) if the sip parser is improved (which is something I'm seriously considering because internally we have tonnes of sipp scripts, all using regex to pull particular bits of headers apart and they're not all quite fully RFC compliant).

wdoekes:
Ooh. I like that extensible `[last_From.value]` suggestion.

pbertera:
What about to use the `:` as a separator instead of `.`  (Eg. `[last_From:value]`)?
The dot can be used into the header names, having an header ending with `.value` isn't maybe so frequent but something that can happen.

PR #258, opened 2016-10-01
Snom patch: header value syntax

pbertera:
**Last header value:** you can extract the header value using the *!* suffix to the `[last_headerName]`

**Example:**
`To: [last_From!]`

will generates a *To:* header containing the last From header value, this is quite useful when you need to send a new request (Eg. re-INVITE into an inbound SIP dialog)

Of course you can do the same using a regexp, but this syntax simplifies a lot the scenario file.

*All the attribution for this patch should go to Snom Technology AG*

### DIFF
--- a/regress/github-header-values/run
+++ b/regress/github-header-values/run
@@ -1,0 +1,22 @@
+#!/bin/sh
+# This regression test is a part of SIPp.
+# Author: Walter Doekes, OSSO B.V., 2016
+. "`dirname "$0"`/../functions"; init
+
+sippfg -m 1 -sf uas.xml -p 5070 \
+    -trace_logs -log_file log.log >/dev/null 2>&1 &
+job=$!
+sippfg -m 1 -sf uac.xml 127.0.0.1:5070 \
+    -timeout 4 -timeout_error >/dev/null 2>&1
+status=$?
+wait $job || status=1
+
+if test $status -eq 0; then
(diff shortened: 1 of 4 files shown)


wdoekes:
Thanks for the patch with testcase.

Could you make that into `[last_From.value]` instead, as suggested in #132 ? The code changes you provided aren't as clean as they could be, but I'll wait with reviewing.

And, this lack all documentation, visible and invisible. A possible way to document its existence could be to add it to one of the built-in scenario's...

(Lastly, your commits aren't atomic changes. They'll need to be squashed before I could merge. And the testcase naming should either be `github-#XXXX` (reference to ticker XXXX on github) or some discriptive name, like `last_header_value` in which case the "github" makes no sense.)

pbertera:
Thanks for the feedback,
- about the naming: I agree that using the `value`keyword will make the syntax more clear, I would like to have an opinion about using `:` as a separator (see #132).
- about the documentation: what about to add a new scenario (uas-req.xml) where the BYE comes from the uas ?

wdoekes:
> I would like to have an opinion about using `:` as a separator

Sounds good to me. (And it provides more legitimacy for the `:` being optional like it is right now.)

> What about to add a new scenario (uas-req.xml) where the BYE comes from the uas?

I'm a bit reluctant to add new scenario's into the binary itself. If you add it to a samples directory as separate xml file, it would work for me. (Especially if a test case runs it, so the scenario's do not get stale.)

wdoekes:
Closes in favor of WIP in #260.

PR #260, opened 2016-10-01
Snom patch: Last_ keywords extension

pbertera:
This patch add the support for the following new scenario keywords:
- **Last_Contact_URI**: returning the last Contact URI value
- **Last_From_Tag**: extracting the last from-tag
- **Last_To_Tag**: extracting the last to-tag

_All the attribution for this patch should go to Snom Technology AG_


wdoekes:
This is not the way to move forward. If you look at the syntax suggested in #132, I would envision these to be called like:

`[last_Contact.uri]`, `[last_From.param.tag]`, `[last_To.param.tag]` or similar.

See https://www.opensips.org/Documentation/Script-Tran-2-2 for ideas.

I don't mind if only ".value", ".uri" and ".param.tag" are implemented, but at least it would provide a solid syntax to build more features on, instead of of adding `E_Message*` for every individually requested feature.

pbertera:
Yes, you are right, I agree with this more general approach instead of adding more `E_Message*`.

Here: https://github.com/SIPp/sipp/compare/master...Snomio:last_extensions I'm working on a solution for this, the patch implements:

* `last_XXX:value`
* `last_XXX:param.tag`
* `last_XXX:uri`

So should cover this and #258, your feedback are appreciated,

Thanks,

pbertera (commented):
Could you look at the code ?
If is sufficiently clean / correct I'll work on the documentation.

Thanks

regress/github-#0260/run
```diff
@@ -0,0 +1,35 @@
...
+            fi
+        else
+            fail "To:uri not found"
+        fi
+    else
+        fail "From header not found"
+    fi
+else
+    fail "process failure"
+fi
```
wdoekes: This could be done without so much nesting with negative tests instead:
```
if test $status -ne 0; then
    fail "process failure"
elif ! grep -q '^LAST From: "Tom Jones" <sip:tom\.jones@wales.uk>;tag=SIPpTag001$' log.log; then
    fail "From header not found"
...
```

regress/github-#0260/uas.xml
```diff
@@ -0,0 +1,51 @@
...
+
+  <recv request="ACK" rrs="true">
+    <action>
+      <log message="LAST Contact: [last_Contact]" />
+      <log message="LAST Via: [last_Via:value]"/>
+      <log message="LAST From: [last_From:value]"/>
+      <log message="LAST To: [last_To:value]"/>
+      <log message="LAST To:uri: [last_To:uri]"/>
+      <log message="LAST Contact:uri: [last_Contact:uri]"/>
+      <log message="LAST From-tag: [last_From:param.tag]"/>
```
wdoekes: Your "key names" don't correspond to the lookups:

Suggest:
`LAST Contact|[last_Contact]`
`LAST Via:value|[last_Via:value]`
etc..

src/call.cpp
```diff
@@ -923,33 +923,72 @@ char * call::get_last_header(const char * name)
...
+                char * ret;
+                if (!(ret = (char *)malloc(strlen(last_tag + 1)))){
+                    ERROR("call::get_last_header: Cannot allocate param.tag !\n");
+                }
+                sprintf(ret, "%s", last_tag);
+                free(last_tag);
+                return ret;
+            }
+            ERROR("Token %s not valid in %s", sep, name);
+        }
```
wdoekes: This function is too big. The code in the if(strcmp..){...} blocks should be delegated to helper methods and duplicate code should be avoided. 

For example:
```
        if (!sep) { /* [last_Header] */
            sprintf(with_colon, "%s:", name);
            return get_header(last_recv_msg, with_colon, false);
        } 

        /* [last_Header:...] */
        sprintf(with_colon, "%.*s", (sep - name + 1), name);
        if (!strcmp(sep, ":")) {
            return get_header(last_recv_msg, with_colon, false);
        }

        char *value = get_header(last_recv_msg, with_colon, true);
        if (!strcmp(sep, ":value")) {
            ; /* done */
        } else if (!strcmp(sep, ":uri")) {
            extract_header_uri(value); /* reuse mem in value (memmove?) */
        } else if (....) {
            ...
        } else {
            ERROR("Token %s not valid in %s", sep, name);
        }

        return value;
```

include/call.hpp
```diff
@@ -291,7 +291,8 @@ class call : virtual public task, virtual public listener, public virtual socket
 
     char * get_header_field_code(const char * msg, const char * code);
     char * get_last_header(const char * name);
-    char * get_last_request_uri();
+    char * get_last_tag (const char *name);
+    char * get_last_request_uri(const char *name);
```
wdoekes: This should not be called `get_last_request_uri` but `get_last_uri_from` or something.

src/call.cpp
```diff
@@ -923,33 +923,72 @@ char * call::get_last_header(const char * name)
...
+        } else {
+            if (!strcmp(sep, ":value")) {
+                snprintf(with_colon, len - 4, "%s", name);
+                return get_header(last_recv_msg, with_colon, true);
+            }
+            if (!strcmp(sep, ":uri")) {
+                snprintf(with_colon, len - 2, "%s", name);
+                char * last_header_uri = get_last_request_uri(with_colon);
+                char * ret;
+                if (!(ret = (char *)malloc(strlen(last_header_uri + 1)))){
```
wdoekes: get_header() returns a static buffer (yuck, I know).

You can't start returning malloc'ed memory all of a sudden: memory leaks.

src/call.cpp
```diff
@@ -972,6 +1011,48 @@ char * call::get_last_request_uri()
     return last_request_uri;
 }
 
+/* Return the last tag from the header line. On any error returns the
+ * empty string.  The caller must free the result. */
+char * call::get_last_tag(const char *name)
```
wdoekes: This should be implemented in `sip_parser.cpp` instead. See `get_peer_tag`.

wdoekes (changes requested):

pbertera:
Thanks for all the hints, can you give me your feedback ?
Thanks a lot,

src/sip_parser.cpp
```diff
@@ -109,6 +109,11 @@ char * get_peer_tag(const char *msg)
     return tag;
 }
 
+char * get_peer_tag(const char *msg)
+{
+    return get_param_tag(msg, "To", "t");
```
wdoekes: This could also be:
```
get_param(msg, "tag", "To", "t");
```
That'd be even more generic.

src/call.cpp
```diff
@@ -2201,7 +2234,7 @@ char* call::createSendingMessage(SendingMessage *src, int P_index, char *msg_buf
             }
             break;
         case E_Message_Last_Request_URI: {
-            char * last_request_uri = get_last_request_uri();
+            char * last_request_uri = get_last_header_uri("To:");
             dest += sprintf(dest, "%s", last_request_uri);
             free(last_request_uri);
```
wdoekes: Why not this?
```
dest += sprintf(dest, "%s", get_last_header("To:uri"));
```
And drop the entire `get_last_header_uri()` method.

The get_uri-code could be implemented in `sip_parser.cpp` instead as `get_first_uri(const char *msg, const char *name, const char *shortname)` or something, with the same semantics as the other get_* functions.

src/call.cpp
```diff
@@ -920,36 +920,69 @@ char * call::get_last_header(const char * name)
...
+        int uri_len = strlen(uri);
+        memmove(value, uri, uri_len);
+        value[uri_len] = '\0';
+        free(uri);
+    } else if (!strcmp(sep, ":param.tag")) { /* [last_Header:param.tag] */
+        snprintf(header_name, len - 9, "%s", name);
+        char * last_tag = get_param_tag(last_recv_msg, header_name, "");
+        int last_tag_len = strlen(last_tag);
+        memmove(value, last_tag, last_tag_len);
+        value[last_tag_len] = '\0';
```
wdoekes:
```
memmove(value, last_tag, strlen(last_tag) + 1);
```
.. copies the NUL along.

src/call.cpp
```diff
@@ -920,36 +920,69 @@ char * call::get_last_header(const char * name)
...
+        return get_header(last_recv_msg, header_name, false);
+    }
+
+    snprintf(header_name, len - strlen(sep) + 2, name);
+
+    /* [last_Header:] */
     if (name[len - 1] == ':') {
         return get_header(last_recv_msg, name, false);
+    }
+    char *value = get_header(last_recv_msg, header_name, true);
```
wdoekes: Apparently you're not using value `get_header(last_recv_msg, header_name, true);` here, move it to the `; /* done */` instead.

It would be only useful here if we used functions that operate on that single header directly later on, e.g.:
* `get_param(value, "tag");` or `extract_param(value, "tag");`
* `get_first_uri(value);` or `extract_first_uri(value);`

They could certainly be implemented and would clean up things some more.

src/call.cpp
```diff
@@ -920,36 +920,69 @@ char * call::get_last_header(const char * name)
...
+
+    if (!strcmp(sep, ":value")) { /* [last_Header:value] */
+        ; /* done */
+    } else if (!strcmp(sep, ":uri")) { /* [last_Header:uri] */
+        char * uri = get_last_header_uri(header_name);
+        int uri_len = strlen(uri);
+        memmove(value, uri, uri_len);
+        value[uri_len] = '\0';
+        free(uri);
```
wdoekes: If get_last_header_uri is dropped, we get rid of this unusual free() here (because of the different semantics of get_last_header_uri).

wdoekes (changes requested):
I'm sorry for being picky. But if we're changing the source, it should be for the better.

pbertera:
@wdoekes no pickiness, unfortunately my programming skills are very weak. Your fine grained review is a warranty for the quality of SIPp, thanks a lot.
I hope the code now is an acceptable level of quality, if yes I'll write an example scenario and I'all add the note to the changelog.
Many thanks,

wdoekes:
Sorry for the delay. I've been busy. I'll try to get to this in the coming week.

wdoekes:
Note to self: this fails:
```
Testing github-#0152: failed (got 0 calls, and 0 failed)
```

PR #278, opened 2017-01-12
Added RTP echo processing for multiple calls with `auto_media_port`

bokrenok:
First approach for correct RTP echo processing for multiple calls.
Added creation of a socket and a thread for each call with auto-incremented RTP port.

Activated by option `-rtp_echo` and by action in script:
Something like this:
    <nop>
        <action>
            <rtp_echo value="1" />
        </action>
    </nop>


wdoekes:
Not sure how I feel about `[auto_media_port]`. There were already too many port settings going on. `[media_port]`, `[rtp_stream_port...]`.

I don't mind the echo feature though. But I wonder if we can do that without even more additional options. Perhaps use a magic `<exec rtpstream>` keyword value instead..?

wdoekes:
> I don't mind the echo feature though. But I wonder if we can do that without even more additional options. Perhaps use a magic <exec rtpstream> keyword value instead..?

Haha. So I merged the rtp_echo already. Okay. That changes things.

PR #279, opened 2017-01-19
Updates to the exec option in sipp scripts.

ltwist:
Changed sip to run exec commands using posix_spawn to improve performance when running a high load of commands. Also added a new option to exec to allow the user to specify whether or not the command should have completed before allowing sipp to continue processing it's script. This stops the sip scheduler resolution from getting out of hand when a high load of commands needs to be ran.


wdoekes:
I'm sorry. This PR consists of 3 different "fixes" spread out over more commits. And it includes a merge failure.

Please do one commit with one fix per PR.

Issue #237, opened 2016-06-13
sipp crash when play_dtmf is used


mansong1:
When I use the exec play_dtmf function within a scenario as attached, sipp crashes: _terminated by signal SIGSEGV (Address boundary error)_ 

I'm running Ubuntu 16.04

[PlayDTMF.zip](https://github.com/SIPp/sipp/files/311515/PlayDTMF.zip)

loxybjorn:
I'm experiencing this too (on a Debian 8 installation)!

wdoekes:
Are you capable of getting a backtrace of the crash? And please specify which version of SIPp you're running.

loxybjorn:
For me its "SIPp v3.6-dev-71-g789c45e-TLS-PCAP-RTPSTREAM built May  9 2016, 12:10:03".

mansong1:
The version I'm using is built of the latest master branch.  I got the following backtrace:

#0  SendingMessage::numComponents (this=this@entry=0x0) at src/message.cpp:599
#1  0x0000000000411c09 in call::createSendingMessage (this=this@entry=0x8b1530, src=0x0, P_index=P_index@entry=-2, 
    msg_buffer=msg_buffer@entry=0x675920 <call::createSendingMessage(SendingMessage*, int, int*)::msg_buffer> "", buf_len=buf_len@entry=65538, msgLen=msgLen@entry=0x0) at src/call.cpp:1900
#2  0x0000000000415ac2 in call::createSendingMessage (msgLen=0x0, P_index=-2, src=<optimised out>, this=0x8b1530) at src/call.cpp:1885
#3  call::executeAction (this=this@entry=0x8b1530, msg=msg@entry=0x0, curmsg=curmsg@entry=0x857410) at src/call.cpp:3683
#4  0x0000000000415f3d in call::executeMessage (this=0x8b1530, curmsg=0x857410) at src/call.cpp:1228
#5  0x000000000040890b in traffic_thread () at src/sipp.cpp:560
#6  main (argc=<optimised out>, argv=<optimised out>) at src/sipp.cpp:1989

vodik:
Fun stuff. I'll see what I can dig out.

loxybjorn:
Here's a backtrace from me:

Program received signal SIGSEGV, Segmentation fault.
SendingMessage::numComponents (this=0x0) at src/message.cpp:599
599         return messageComponents.size();
(gdb) bt
[...]

PR #285, opened 2017-02-09
Fix for Bug#237 to stop a crash when play_dtmf is used.

aaicken:
Two changes:

1. Fix for [Bug#237](https://github.com/SIPp/sipp/issues/237).
2. Also modified gitignore to ignore core files as I ended up with a core file in my working directory.

Issue 237 was first replicated, causing sipp to crash and produce a core file. The core file was debugged to trace back to the root cause of the crash. The scenario was re-tested with the modified code and it now runs ok and RFC2833 DTMF events are being generated.

### DIFF
--- a/src/scenario.cpp
+++ b/src/scenario.cpp
@@ -1617,7 +1617,7 @@ void scenario::parseAction(CActions *actions)
                 hasMedia = 1;
                 free(ptr);
             } else if ((cptr = xp_get_value("play_dtmf"))) {
-                tmpAction->setMessage(ptr);
+                tmpAction->setMessage(cptr);
                 tmpAction->setActionType(CAction::E_AT_PLAY_DTMF);
                 hasMedia = 1;
 #else


.gitignore
```diff
@@ -16,6 +16,7 @@ config.log
 config.status
 config.sub
 compile
+core
```
wdoekes: We don't need to ignore core files. They're not part of the standard SIPp compilation cycle, I hope ;P

Please drop this from the changeset.

src/scenario.cpp
```diff
@@ -1617,7 +1617,7 @@ void scenario::parseAction(CActions *actions)
                 hasMedia = 1;
                 free(ptr);
             } else if ((cptr = xp_get_value("play_dtmf"))) {
-                tmpAction->setMessage(ptr);
+                tmpAction->setMessage(cptr);
```
wdoekes: Good find. Was indeed missed in 04ba8c59 change.

Thanks!

wdoekes:
Cool.

Could you just squash that into a single commit and force push? That will update this PR with the single commit.

Thanks!

PR #296, opened 2017-03-22
docs reformating and xml highlighting

nshopik:
This should cover 90% docs now. Only missing table is `List of commands with their attributes` and XML scenario files aren't imported and thus links aren't correct. These produce errors and warrings during `make html` but otherwise its fine

You are welcome to read-prof docs and make changes, like removing outdated info or add new. Only major changes I plan for `ownscenarios.rst` otherwise other files need just some face lifting and minor fixes


docs/scenarios/actions.rst
```diff
@@ -42,33 +42,47 @@ Here is the syntax of the regexp action:
...
+                          SIPp string variable). 
+header            None    Header to try to match against.
+                          Only used when the search_in tag is set to hdr. MANDATORY IF search_in
+                          is equal to hdr. 
+variable          None    Variable to try to match against. Only
+                          used when the search_in tag is set to var. MANDATORY IF search_in is
+                          equal to var. 
+case_indep        false   To look for a header ignoring case .
+                          Only used when the search_in tag is set to hdr. 
+occurence     1           To find the nth occurence of a header. Only used when the search_in tag is set
```
wdoekes: Has been fixed to `occurrence` (double-r). Althrough single-r is still legal for backwards compatibility.

nshopik: Well I guess i'll change it in next big commit then, to avoid bloating commits history

PR #299, opened 2017-04-10
Convert to FlatTables

nshopik:
This should update latest 2 big tables


docs/scenarios/actions.rst
```diff
@@ -60,7 +60,7 @@ variable          None    Variable to try to match against. Only
                           equal to var. 
 case_indep        false   To look for a header ignoring case .
                           Only used when the search_in tag is set to hdr. 
-occurence     1           To find the nth occurence of a header. Only used when the search_in tag is set
+occurence         1       To find the nth occurence of a header. Only used when the search_in tag is set
```
wdoekes: "occurrence" (double-r, twice)

It has been fixed in the code, but the old misspelling is also allowed.

nshopik:
squashed merge commit but not sure if i made it right it still show changes from previous commit, fixed typo

wdoekes:
Ah, yes.

Instead of squashing the merge, you should've rebased. (You should still be able to do that.)

    git fetch upstream
    git rebase upstream/master

nshopik:
yay, thanks for pointers. Its should be safe to merge.

So far only missing pieces are integrated scenarios and links across documentation.  We should mark SF documentation as deprecated/outdated as no changes will be made to it

wdoekes:
Hm. Style is now broken/gone:
https://sipp-wip.readthedocs.io/en/latest/

nshopik:
Yeah I've noticed too, but on my local installation it was fine, there is `theme_overrides.css` which may cause that. But I copy-pasted it from kernel docs, as I don't know anything about css. It suppose to make wider view to make room for big tables

nshopik:
Looks like its missing main CSS ` <link rel="stylesheet" href="_static/css/theme.css" type="text/css" />` when I enable overide but local build does include main + override when building

Issue #302, opened 2017-04-27
configure: error: ssl library missing


seba2044:
Hi 
Arch linux, openssl is installed properly.
I've got bug when I try to use ./build.sh --full and ./configure --with-pcap --with-sctp --with-openssl. 
```
checking for SSL_library_init in -lssl... no

configure: error: ssl library missing
```
I don't understand ssl check in configure script (line #7363) but I made my own which works fine.  

```
#include <openssl/ssl.h>

int
main ()
{
        printf("%d\n"SSL_library_init ());
        return 0;
}
```

`gcc cos.c -lssl`

So I assume that error is in config script.

wdoekes:
Methinks it needs a few new-ssl fixes, like these:
https://github.com/wdoekes/pam-imap/commit/3d9c164ed32644a3da2b37cadec7ac93758a49d1

Thanks for the report!

seba2044:
`src/socket.cpp: In function ‘ssl_init_status FI_init_ssl_context()’:
src/socket.cpp:2101:36: warning: ‘const SSL_METHOD* TLSv1_method()’ is deprecated
[-Wdeprecated-declarations]
sip_trp_ssl_ctx = SSL_CTX_new( TLSv1_method() );
^~~~~~~~~~~~
In file included from
/usr/include/openssl/ct.h:13:0,
        from
        /usr/include/openssl/ssl.h:61,
        from
        ./include/sslcommon.h:25,
        from
        ./include/socket.hpp:24,
        from
        ./include/sipp.hpp:75,
        from
        src/socket.cpp:41:
        /usr/include/openssl/ssl.h:1596:1:
        note:
        declared
        here
        DEPRECATEDIN_1_1_0(__owur
                        const
                        SSL_METHOD
                        *TLSv1_method(void))
        /* TLSv1.0
         * */
        ^
        src/socket.cpp:2101:49:
        warning:
        ‘const
        SSL_METHOD*
        TLSv1_method()’
        is
        deprecated
        [-Wdeprecated-declarations]
        sip_trp_ssl_ctx
        =
        SSL_CTX_new(
                        TLSv1_method()
                   );
        ^
        In
        file
        included
        from
        /usr/include/openssl/ct.h:13:0,
        from
        /usr/include/openssl/ssl.h:61,
        from
        ./include/sslcommon.h:25,
        from
        ./include/socket.hpp:24,
        from
        ./include/sipp.hpp:75,
        from
        src/socket.cpp:41:
[...]

PR #304, opened 2017-04-28
Fix problem with openssl 1.1

seba2044:
Do I fix it correctly ? #302


wdoekes:
Good try. But you've broken support for SSL on older systems:

https://travis-ci.org/SIPp/sipp/jobs/226818690#L502

PR #308, opened 2017-05-19
Add initial xml schema  RELAX NG

fleith:
Contain the most basic syntax, but need some improvements on advanced scenarios.

### DIFF
--- a/sipp.rnc
+++ b/sipp.rnc
@@ -1,0 +1,192 @@
+default namespace = ""
+
+start =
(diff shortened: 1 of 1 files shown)


wdoekes:
I'm sorry. I don't do relax-ng, and if I merge this, it can only go out of sync because I wouldn't be updating it.

You'll have to make a more compelling case to get this included. And fix something so it stays in sync with the codebase.

Thanks for the PR though.

Issue #311, opened 2017-06-05
Add 'check_it' and 'check_it_inverse' attributes to 'test' and 'strcmp' tags


fadimko:
It would be nice to have these attributes for automated testing.

PR #312, opened 2017-06-05
Add 'check_it' and 'check_it_inverse' attributes to 'test' and 'strcmp' tags

fadimko:
#311


fadimko:
I also have the branch with the patch for the current stable version 3.5.1. So if this pr will be merged, may be we can have 3.5.2 release?

jdesmarais-jive:
Exactly what I was looking for. 👍

wdoekes:
This is not going into 3.5 (it adds features and doesn't fix problematic bugs). But merging to 3.6 now.

PR #317, opened 2017-07-08
Enable SiPp to use TLS1.2

arunrajm:


orgads:
:+1:

orgads:
Any reason not to accept this?

wdoekes:
The only reason would be backwards compatibility. But it does pass the tests, so that's good.

In short: it's probably going in.

wdoekes:
> TLS_method(), TLS_server_method(), TLS_client_method()
> These are the general-purpose version-flexible SSL/TLS methods. The actual protocol version used will be negotiated to the highest version mutually supported by the client and the server. The supported protocols are TLSv1, TLSv1.1 and TLSv1.2. Applications should use these methods and avoid the version-specific methods described below.

I'm guessing those should be appropriate instead. (Note that we're not going for best-encryption here, but for compatible with the most devices.)

wdoekes:
That would be something for autoconf then.

orgads:
Proposed a new patch: https://github.com/SIPp/sipp/pull/341

wdoekes:
Fixed by 553290318. Thanks!

PR #326, opened 2017-10-18
Support escaped brackets inside messages

groegeorg:
Sometimes it's necessary to add square brackets inside a message.
Anything inside brackets is normally treated as a keyword. With this
commit it's possible to escape brackets with a leading backslash.

### DIFF
--- a/src/message.cpp
+++ b/src/message.cpp
@@ -154,6 +154,12 @@ SendingMessage::SendingMessage(scenario* msg_scenario, const char* const_src, bo
                 *dest++ = val & 0xff;
             }
             src++;
+        } else if ((*src == '\\') && (*(src+1) == '[')) {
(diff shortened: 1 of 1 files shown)


wdoekes:
Why don't you use `\x5b` and `\x5d`? It's documented 11 lines above your edit:

`/* Allows any hex coded char like '\x5B' ([) */`

If we're adding `\[`, we must also add `\\`. And in the mean time, we'd break everyones script who intentionally adds a single backslash (e.g. `\[callid]` where they want `\CALLID`).

I think we can work with just using the exising `\xNN` and leaving the parser as is.

I'm open to documentation improvements though.

groegeorg:
Alright, I didn't know about the `\xNN` escapes. I'll take the pull request back then!

PR #335, opened 2018-01-29
SSL: separate CA file

bluerise:
This commit allows using a separate file to specify the CA to trust.
This way we have two files (cert, key) for our own certificate and
one file with the trusted CA.  Putting the own certificate and the CA
in the same file, as previously done with the tls_cert option, does
not seem to work.


fenichelar:
@bluerise This is the way NGINX, Apache, and everything else I have ever used works..... It is so much more clear. +1

PR #340, opened 2018-04-22
TLS: Do not exit on SSL connection failure

orgads:
A single failed socket should not fail the entire test.

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -1573,8 +1573,9 @@ int SIPpSocket::connect(struct sockaddr_storage* dest)
                 sipp_usleep(SIPP_SSL_RETRY_TIMEOUT);
                 continue;
             }
-            ERROR("Error in SSL connection: %s\n", SSL_error_string(err, rc));
-            break;
+            WARNING("Error in SSL connection: %s\n", SSL_error_string(err, rc));
+            invalidate();
+            return err;
         }
 #else
         ERROR("You need to compile SIPp with TLS support");


orgads:
ping?

wdoekes:
Hi. I'm trying to find the time to do a new release. You have been noticed. Thanks for the PR and the ping :)

orgads:
I'd really like to see this change in the new release. Do you find it risky?

wdoekes:
> Do you find it risky?

Yes. I don't use the TLS code, and I don't know off-hand what happens when we invalidate() a socket. (Will it cause double frees?)

I looked at the code now, and I don't see any problem with invalidate() per se, but that leaves me with the question what happens when I do a single run (`-m 1`). I need to be sure that in that case failed connection will still cause SIPp to exit with a non-zero status.

What is the ssl error you are encountering?

orgads:
We tested a high load of TLS connections with SIPp, and some of the calls reached handshake timeout.

We can test another type of failure (mismatch of cipher suite) with -m1. But generally, do you see a reason why an entire execution should fail because of a single connection failure?

wdoekes:
Okay. Should be good. I'll go ahead and merge this.

Thanks for your patience :)

orgads:
ping?

orgads:
PING!

orgads:
Hello?

pamalpass:
Is it me you’re looking for?

wdoekes:
Sorry for the long wait. Thanks!

orgads:
Thank you.

PR #341, opened 2018-04-23
TLS: Use highest negotiated protocol version

orgads:
For OpenSSL 1.0.x this will be v1.2. For 1.1.x it will be chosen by
negotiation.


orgads:
Tested compilation with both openssl 1.0.2 and 1.1.0.

PR #344, opened 2018-05-08
Retry video RTP bind if port is taken.

### LINKED ISSUE #276: Fetching media_port fails if media_port+2 is unavailable
If a port is found for audio, the video port is taken as port+2. If that is unavailable, things break fast

https://github.com/SIPp/sipp/blob/v3.5.1/src/sipp.cpp#L2184..L2202
^-- tries 100 times

https://github.com/SIPp/sipp/blob/v3.5.1/src/sipp.cpp#L2213..L2230
^-- simply takes audio_port+2

coreyfarrell:
Fixes #276


wdoekes:
Thanks for the patch!

Not so fond of the commando pattern with the goto and +=3 though. So I rewrote it a bit. Please try c1158cf564e9117cb716301082e0d409ad654bfd (fix/3.5/retry-video-port-276).

coreyfarrell:
@wdoekes updated patch works for me and I agree it looks better without the goto.  Do you want me to close this PR or are you going to update it?

PR #348, opened 2018-05-11
Fix traffic thread crash upon shutdown

jeannotlanglois:
Commit f2cf583b intended to remove a double free upon shutdown but ended up causing a crash - this attempts to fix the problem.


jeannotlanglois:
Need to abort submission to add missing NULL check...

PR #349, opened 2018-05-11
Fix traffic thread crash upon shutdown (introduced by f2cf583b)

jeannotlanglois:
Commit f2cf583b intended to remove a double free upon shutdown but ended up causing a crash when the return value of "traffic_thread()" is false (e.g. shutdown) - this fixes the issue by restoring the original behavior while also using recent changes.

### DIFF
--- a/src/sipp.cpp
+++ b/src/sipp.cpp
@@ -1991,15 +1991,15 @@ int main(int argc, char *argv[])
         }
     }
 
-    traffic_thread();
-
-    /* Clean up */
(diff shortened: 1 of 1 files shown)


jeannotlanglois:
The Travis CI "failure" appears to be specific to the OSX build not working - I have doubts that this has anything to do with my pull request.

wdoekes:
Thanks!

jeannotlanglois:
Thanks -- glad my very first github submission is okay :)

PR #350, opened 2018-05-16
performance enhancements with warning removal

zhouyu8x:


wdoekes:
Hi there. It looks sane, so thanks!

Do you want to elaborate on any of the changes?

PR #354, opened 2018-07-04
Fix TCP binding regression (introduced by 337f5c0c)

jeannotlanglois:
Commit 337f5c0c caused TCP binding to no longer work on a given
interface (specified by the "-i w.x.y.z" parameter) when
the "-p <port>" commandline parameter is present.

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -2568,16 +2568,6 @@ int open_connections()
                            sizeof (sock_opt)) == -1) {
                 ERROR_NO("setsockopt(SO_REUSEADDR) failed");
             }
-
-            if (tcp_multiplex->ss_ipv6) {
-                struct sockaddr_in6 sa_loc = {AF_INET6};
-                sa_loc.sin6_port = htons(local_port);
-                sipp_bind_socket(tcp_multiplex, (struct sockaddr_storage*)&sa_loc, NULL);
-            } else {
-                struct sockaddr_in sa_loc = {AF_INET};
-                sa_loc.sin_port = htons(local_port);
-                sipp_bind_socket(tcp_multiplex, (struct sockaddr_storage*)&sa_loc, NULL);
-            }
         }
 
         if (tcp_multiplex->connect(&remote_sockaddr)) {


jeannotlanglois:
@wdoekes 

Can this change be merged?  It is a bad regression that has been around since 2015... it should be fixed (affects both TCP+TLS scenarios).

jeannotlanglois:
@wdoekes : Any updates on the merging of this fix for 3.6-dev?

jeannotlanglois:
@wdoekes :  Thanks a bunch for looking into this.  So today I've spent my time testing your change with a lot of XML scenarios I have (I'm using two instances of SIPP as part of a big test harness script to test a B2BUA gateway).  

There is one good news and one bad news with your overall changes to my initial submission:

1) Good news -- the TCP/TLS binding issue has disappeared -- I've definitely tested both intensively with my test harness scenarios.  I've also tested UDP with my test harness while I was at it even though it is not related (works fine too).

2) Bad news -- it appears that a recent commit you made, "1a98eb19" (related to a supposed premature epoll close), has created a crash under specific conditions when using the TLS transport ("-t l1") as well as the "-trace_msg" parameter with XML scenarios that do SIP forking.  It took me quite a while to isolate the scenario and the commit, but I've definitely pinned it down.  If I locally revert commit "1a98eb19" the crash no longer happens.  I'm not too familiar with the signalling socket code in the main() of SIPP but there's something that appears to require "ctrl_socket->close()" to be done _before_ "free(epollevents);" -- otherwise a crash ensues.  This is a distinct separate issue from the TCP/TLS binding one.

Applying this appears to resolve the problem for me:
diff --git a/src/sipp.cpp b/src/sipp.cpp
index 641c628..03b8f50 100644
--- a/src/sipp.cpp
+++ b/src/sipp.cpp
@@ -2024,16 +2024,16 @@ int main(int argc, char *argv[])
         pthread_join(pthread3_id, NULL);
     }

-    if (ctrl_socket != NULL) {
-        ctrl_socket->close(); /* also uses epoll */
-        ctrl_socket = NULL; /* close "delete's this" */
-    }
-
 #ifdef HAVE_EPOLL
     close(epollfd);
     free(epollevents);
 #endif

     free(scenario_file);
+    if (ctrl_socket != NULL) {
+        ctrl_socket->close(); /* also uses epoll */
+        ctrl_socket = NULL; /* close "delete's this" */
+    }
+
     sipp_exit(EXIT_TEST_RES_UNKNOWN);
 }

I'm working on an extension of the RTPSTREAM code for a new SIPP feature which I hope to submit for a later version of SIPP.  It's possible that my local changes have something that requires ctrl_socket to be closed after epollevents is freed -- and it might be something sneaky.

Please let me know if you'd like me to open a bug for this separate issue.

Thanks!

wdoekes:
I added the problem description to the commit here in github:
https://github.com/SIPp/sipp/commit/1a98eb190c12304a77d30001609d13df5f372727

Closing the control socket means running the epoll unregister code. That cannot be done *after* the closing of epollfd itself.

However, in all cases, the closing of stdin is wrong. We should not do this, as it causes breakage in an interactive terminal. Nevertheless I'm not fond of the (ctrl) socket instance being left unfreed there either.

This should fix the terminal problem:
```
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -51,6 +51,9 @@ extern bool show_index;
 SIPpSocket *ctrl_socket = NULL;
 SIPpSocket *stdin_socket = NULL;
 
+static int stdin_fileno = -1;
+static int stdin_mode;
+
 /******************** Recv Poll Processing *********************/
 
 unsigned pollnfds;
@@ -541,11 +544,19 @@ void setup_ctrl_socket()
     }
 }
 
+void reset_stdin()
+{
+    fcntl(stdin_fileno, F_SETFL, stdin_mode);
+}
+
 void setup_stdin_socket()
 {
-    int stdin_mode = fcntl(fileno(stdin), F_GETFL);
-    fcntl(fileno(stdin), F_SETFL, stdin_mode | O_NONBLOCK);
-    stdin_socket = new SIPpSocket(0, T_UDP, fileno(stdin), 0);
+    stdin_fileno = fileno(stdin);
+    stdin_mode = fcntl(stdin_fileno, F_GETFL);
+    atexit(reset_stdin);
+    fcntl(stdin_fileno, F_SETFL, stdin_mode | O_NONBLOCK);
+
+    stdin_socket = new SIPpSocket(0, T_UDP, stdin_fileno, 0);
     if (!stdin_socket) {
         ERROR_NO("Could not setup keyboard (stdin) socket!\n");
     }
@@ -946,12 +957,12 @@ void SIPpSocket::invalidate()
         }
 #endif
 
-        if (::close(ss_fd) < 0) {
+        if (ss_fd == stdin_fileno) {
+            /* don't close stdin, breaks interactive terminals */
+        } else if (::close(ss_fd) < 0) {
             WARNING_NO("Failed to close socket %d", ss_fd);
         }
-        else {
-            ss_fd = -1;
-        }
+        ss_fd = -1;
     }
 
     if ((pollidx = ss_pollidx) >= pollnfds) {
```
But I don't think it will fix any crashes of yours.

Why would you need the ctrl-socket *after* closing epoll and being done with all the threads? I think you'll need to investigate that instead of reordering the shutdown code. This is likely related to the problems you where having when you submitted 83a9ddea. Are you spawning more threads that are still running when we're here?

And yes, this should've been a separate issue.

wdoekes:
So, I can reproduce. That helps.

jeannotlanglois:
> Okay, please try https://github.com/SIPp/sipp/tree/fix/shutdown-issues-354

Will do.  I will post further comments under issue #359 to keep things organized for the separate ctrl_socket shutdown problem.

PR #356, opened 2018-07-04
Implement "-tls_version <x>" parameter to select which TLS protocol v…

jeannotlanglois:
…ersion to use (1.0, 1.1 or 1.2)

Commit "55329031" changed the behavior of TLS_init_context()
to always use TLS v1.2.  However when using TLS with non-SIPP
endpoints this can cause issues if they do not support v1.2.

This change allows explicitly specifying which TLS protocol
version to use on the commandline (default is 1.2).

### DIFF
--- a/src/sipp.cpp
+++ b/src/sipp.cpp
@@ -165,6 +165,7 @@ struct sipp_option options_table[] = {
     {"tls_cert", "Set the name for TLS Certificate file. Default is 'cacert.pem", SIPP_OPTION_STRING, &tls_cert_name, 1},
     {"tls_key", "Set the name for TLS Private Key file. Default is 'cakey.pem'", SIPP_OPTION_STRING, &tls_key_name, 1},
     {"tls_crl", "Set the name for Certificate Revocation List file. If not specified, X509 CRL is not activated.", SIPP_OPTION_STRING, &tls_crl_name, 1},
+    {"tls_version", "Set the TLS protocol version to use (1.0, 1.1, 1.2) -- default is 1.2", SIPP_OPTION_FLOAT, &tls_version, 1},
 #else
     {"tls_cert", NULL, SIPP_OPTION_NEED_SSL, NULL, 1},
     {"tls_key", NULL, SIPP_OPTION_NEED_SSL, NULL, 1},
--- a/src/sipp.cpp
+++ b/src/sipp.cpp
@@ -170,6 +170,7 @@ struct sipp_option options_table[] = {
     {"tls_cert", NULL, SIPP_OPTION_NEED_SSL, NULL, 1},
     {"tls_key", NULL, SIPP_OPTION_NEED_SSL, NULL, 1},
     {"tls_crl", NULL, SIPP_OPTION_NEED_SSL, NULL, 1},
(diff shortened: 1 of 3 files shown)


src/sipp.cpp
```diff
@@ -165,6 +165,7 @@ struct sipp_option options_table[] = {
     {"tls_cert", "Set the name for TLS Certificate file. Default is 'cacert.pem", SIPP_OPTION_STRING, &tls_cert_name, 1},
     {"tls_key", "Set the name for TLS Private Key file. Default is 'cakey.pem'", SIPP_OPTION_STRING, &tls_key_name, 1},
     {"tls_crl", "Set the name for Certificate Revocation List file. If not specified, X509 CRL is not activated.", SIPP_OPTION_STRING, &tls_crl_name, 1},
+    {"tls_version", "Set the TLS protocol version to use (1.0, 1.1, 1.2) -- default is 1.2", SIPP_OPTION_FLOAT, &tls_version, 1},
 #else
     {"tls_cert", NULL, SIPP_OPTION_NEED_SSL, NULL, 1},
     {"tls_key", NULL, SIPP_OPTION_NEED_SSL, NULL, 1},
```
wdoekes: You'll want a tls_version in this `#else` block as well.

src/sslsocket.cpp
```diff
@@ -190,15 +186,35 @@ static int sip_tls_load_crls(SSL_CTX* ctx , const char* crlfile)
...
+        sip_trp_ssl_ctx_client = SSL_CTX_new(TLSv1_method());
+    } else if (tls_version == 1.1) {
+        sip_trp_ssl_ctx_client = SSL_CTX_new(TLSv1_1_method());
+    } else if (tls_version == 1.2) {
+        sip_trp_ssl_ctx_client = SSL_CTX_new(TLSv1_2_method());
+    } else {
+        ERROR("Unrecognized TLS version for client context: %1.1f\n", tls_version);
+        sip_trp_ssl_ctx_client = NULL;
+    }
+
```
wdoekes: Duplicate code.

Perhaps refactor the tls_version `if` into a static function?

jeannotlanglois:
@wdoekes 

>You'll want a tls_version in this #else block as well.
Oh yes - I had not realized that.  Indeed.  Fixed.

>Duplicate code. Perhaps refactor the tls_version if into a static function?
Yes - agreed.  Refactored into helper static function.

I've re-ran all my testcases (and also validated the 'else' case where no TLS support is built in).  

I've resubmitted my changes as a new commit (I was initially thinking of squashing everything into a single commit to make git history cleaner and facilitate merging but since I've started this pull request I guess this might not be a good idea as it could break things if I deleted the branch to recreate it afterwards).

It appears that one of the Travis CI testcases stalled for a long time - just for one of the builds (not sure why, wondering if this is a bad coincidence since it's only one of the builds).  I do not know enough to troubleshoot (see https://travis-ci.org/SIPp/sipp/jobs/400449783).

Let me know if there's something that still requires changing.

wdoekes:
Re-ran the test and it completed just fine :+1:

wdoekes:
Thanks!

jeannotlanglois:
Thanks! :)

PR #358, opened 2018-07-08
Add playback file name option

piotrgregor:
Can be used from the commandline to pass name of the file
to be used as a playback in scenario for "exec play_pcap_audio"
and "exec rtp_stream" actions:

    <exec play_pcap_audio="pcap/test1.pcap"/>
    <exec rtp_stream="pcap/test2.raw"/>

Usage:

    sipp -playback_file pcap/test3.pcap (will play test3.pcap instead of test1/2.pcap).

This way the same scenario file can be used for playing different playback files
without the need to change the scenario file.

### DIFF
--- a/include/sipp.hpp
+++ b/include/sipp.hpp
@@ -321,6 +321,8 @@ extern double             tls_version             _DEFVAL(1.2);
 extern char*              scenario_file           _DEFVAL(NULL);
 extern_c char*            scenario_path           _DEFVAL(NULL);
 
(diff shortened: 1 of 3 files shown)


wdoekes:
I don't think this is the way to go.

I'd rather see a `[replaceable]` key to be used there.

```
<exec play_pcap_audio="[pcap_file"/>

wdoekes:
I like #330 better! Thanks for the PR though.

PR #366, opened 2018-09-06
Add media support for G722 and iLBC codecs

hafkensite:
Added the required constants for G722 (payload 9) and iLBC at 30ms per frame to the media actions.


wdoekes:
Thanks!

PR #370, opened 2018-11-10
Fixed crash when exiting thread on sendto() failure

treyadg:
pthread_cleanup_pop() should be called after calling pthread_cleanup_push(), in case of sendto() failure, the function will return without calling pthread_cleanup_pop() which will cause sipp to crash with segfault.


wdoekes:
Thanks! I'll add a `return (ret >= 0 ? 0 : -1)` for good measure.

PR #400, opened 2019-05-24
fixed compiling bug on cygwin

Vasya-M:
I get follow error:
src/socket.cpp:758:14: error: ‘strcasestr’ was not declared in this scope
     if ((l = strcasestr(socketbuf->buf + socketbuf->offset, "\r\nContent-Length:"))) {
              ^~~~~~~~~~
src/socket.cpp:758:14: note: suggested alternative: ‘strcasecmp’
     if ((l = strcasestr(socketbuf->buf + socketbuf->offset, "\r\nContent-Length:"))) {
              ^~~~~~~~~~
              strcasecmp

### DIFF
--- a/include/sip_parser.hpp
+++ b/include/sip_parser.hpp
@@ -23,6 +23,9 @@
 #ifndef __SIPP_SIP_PARSER_H__
 #define __SIPP_SIP_PARSER_H__
 
+#include <ctype.h>
+#include <string.h>
+
 #define MAX_HEADER_LEN 2049
 
 char *get_call_id(const char* msg);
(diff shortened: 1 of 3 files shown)


include/sip_parser.hpp
```diff
@@ -23,6 +23,9 @@
 #ifndef __SIPP_SIP_PARSER_H__
 #define __SIPP_SIP_PARSER_H__
 
+#include <ctype.h>
+#include <string.h>
```
wdoekes: This is not needed here. If it is needed, it should be in src/sip_parser.cpp.

src/sip_parser.cpp
```diff
@@ -160,7 +160,7 @@ char* get_header(const char* message, const char* name, bool content)
         snprintf(header_with_newline, MAX_HEADER_LEN, "\n%s", name);
         dest = last_header;
 
-        while ((src = strcasestr(src, header_with_newline))) {
+        while ((src = strstr(src, header_with_newline))) {
```
wdoekes: You can't just make a case insensitive search into a sensitive one.

wdoekes:
Please try c98d54638

PR #405, opened 2019-06-06
Play wav file more than 8 bits 

dagueh:
Hello, 
Is that possible to add the possibility to play wav audio file with more than 8 bits.
It's really hard to find ( Maybe for me) a way to encode PCM 8bits audio files.
Best Regards,


wdoekes:
Please don't open crazy pull requests. Use the issues for RFEs.

PR #416, opened 2019-09-09
scenario.xml

PilnikAndrey:


rkday:
I don't really understand what this is for or what the value of merging it would be - feel free to resubmit with a clearer pull request title and a description explaining the benefit, but for now I'm going to close.

PR #423, opened 2019-10-20
Improvements to <exec> function

### LINKED ISSUE #7: Improvements to <exec> function
SIPped, the Polycom fork of SIPp (https://github.com/SIPp/polycom-sipped) has improved the <exec> command, adding the ability to wait for a return code and verify it, as well as output tracing. It would be good to get this merged into the main SIPp code.

It looks as though the relevant commits are 9013201, bca0bcc and 2a23974 for return code verification (which might be a good task for a new dev) and 9013201, 571793a, f7d2e97, 1c9f029 and ba9f6b7 for improved tracing.

dolk13:
I added only 'exec verify'. 
Improving output tracing brings a lot of changes. I will try to do PR later.

resolve #7

### DIFF
--- a/include/actions.hpp
+++ b/include/actions.hpp
@@ -57,6 +57,7 @@ public:
         E_AT_LOG_WARNING,
         E_AT_LOG_ERROR,
         E_AT_EXECUTE_CMD,
+        E_AT_VERIFY_CMD,
         E_AT_EXEC_INTCMD,
         E_AT_VAR_ADD,
         E_AT_VAR_SUBTRACT,
--- a/src/actions.cpp
+++ b/src/actions.cpp
@@ -111,6 +111,8 @@ void CAction::afficheInfo()
         }
     } else if (M_action == E_AT_EXECUTE_CMD) {
         printf("Type[%d] - command[%-32.32s]", M_action, M_message_str[0]);
+    } else if (M_action == E_AT_VERIFY_CMD) {
+        printf("Type[%d] - verify[%-32.32s]", M_action, M_message_str[0]);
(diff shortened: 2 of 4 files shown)


wdoekes:
@rkday : something you want to review?

This may need some docs, as documenting in code isn't enough it people are expected to find it.

rkday (changes requested):
@dolk13 , thanks for this!

As Walter says, the docs need an update to mention `verify`:

- https://github.com/SIPp/sipp/blob/master/docs/scenarios/actions.rst#external-commands
- and a DTD update at https://github.com/SIPp/sipp/blob/master/sipp.dtd#L102

Probably the key thing to mention as a difference between `exec command` and `exec verify` is that the latter is synchronous rather than asynchronous - SIPp will wait for the verified command to run before proceeding, and nothing can happen (no SIP messages can be handled, new calls set up, etc.) while it is running.

That makes me somewhat nervous, because I think this implementation is only suitable for very low loads (or very fast commands) - I ran a scenario with a verify-command that took 200ms to run, and despite me requesting a call setup rate of 10/sec it only actually achieved 2.7/sec.

I think a better approach would be:

- don't call waitpid() in call.cpp
- instead, pause the call (task.cpp has a setPaused method) and add a pointer to the call to a std::map keyed off the PID
- at the start of each iteration in traffic_thread in sipp.cpp, call waitpid with the WNOHANG option and a PID of -1, to check of any child process has exited
- if it has, unpause the call associated with it, and verify the return code

But I think I'd be OK to accept this without this change as long as the impact on the call rate was clearly documented, and improve performance in future.

PR #425, opened 2019-10-22
Update dtd

leedm777:
Ran into some issues using the current DTD to write custom scenarios. There were enough errors and confusing organization that I rewrote it.


rkday:
Thanks! I've merged this and also raised https://github.com/SIPp/sipp/pull/432, which validates all .xml files in the source tree against the DTD as part of our CI, which should keep this working in future.

(I've also made one further fix to the DTD - `nop` can accept a `display` attribute.)

PR #428, opened 2019-10-25
Capture rtp host and port from sdp

Marc-i2x:
This allows the creation of RTP streams with ffmpeg instead of pcap replay. Snippet from UAC scenario:
```xml
  <!-- Play RTP stream through ffmpeg   -->
  <nop>
    <action>
      <exec command="ffmpeg -nostats -loglevel 0 -re -i ~/arbitrary.wav -ar 8000 -acodec pcm_mulaw -f rtp rtp://[last_media_ip]:[last_media_port] >>/dev/null" />
    </action>
  </nop>
```


wdoekes:
Hi!

Could you rebase this against master and fix so all tests succeed?

Thanks!

Marc-i2x:
> Could you rebase this against master and fix so all tests succeed?

done. I also made find_in_sdp availability independent of compile flags as the PR will allow sending RTP without pcap support.

wdoekes:
Interesting idea. I'm a bit reluctant to allow overloading the last_HEADER for this purpose though, as media_ip is not a header (obviously).

It's already a jungle of small little hacks here and there.

How about you make it `[lastsdp_media_ip]` instead. Add the appropriate docs, so people will actually know the feature exists. And indent with 4 spaces (*). Thanks! :)

(*) See: https://github.com/SIPp/sipp/commit/7e62db1ca609ad786384eda3d381019b9ec93a31 and
`astyle --indent=spaces=4 --unpad-paren --pad-header --pad-comma --pad-oper ...`

rkday:
Or perhaps `peer_media_ip`, `peer_media_port`, by analogy with `peer_tag_param`?

src/call.cpp
```diff
@@ -957,6 +955,19 @@ char * call::get_last_header(const char * name)
         ERROR("call::get_last_header: Header to parse bigger than %d (%zu)", MAX_HEADER_LEN, strlen(name));
     }
 
+    if(strcmp(name, "media_ip")==0) {
```
rkday: I wonder if it might be slightly better to always call `extract_rtp_remote_addr` (at line 2820 or so) instead of guarding it with `#ifdef RTP_STREAM`, and store the host/port off as member variables on the `call` object, then just use those (in the same way we use the pre-stored `peer_tag` when handling `E_Message_Peer_Tag_Param`). That would guarantee that the host/port would be retrieved consistently whether you used built-in RTP streaming or this method.

Marc-i2x:
> Interesting idea. I'm a bit reluctant to allow overloading the last_HEADER for this purpose though, as media_ip is not a header (obviously).
> 
> It's already a jungle of small little hacks here and there.
> 
> How about you make it `[lastsdp_media_ip]` instead. 
Could you guide me how to achieve this? Where would I have to add the respective code?

> Add the appropriate docs, so people will actually know the feature exists. 
Will do once I find some free time

> And indent with 4 spaces (*). Thanks! :)
Will do

> 
> (*) See: [7e62db1](https://github.com/SIPp/sipp/commit/7e62db1ca609ad786384eda3d381019b9ec93a31) and
> `astyle --indent=spaces=4 --unpad-paren --pad-header --pad-comma --pad-oper ...`

PR #429, opened 2019-10-26
Use CMake to simplify the build, and build a fully static binary with Github Actions

rkday:
This uses CMake, replacing ~500 lines of automake code with ~125.

This also gives the ability to do a static build, creating a ~9MB binary with no external dependencies that will run anywhere ([sipp.gz](https://github.com/SIPp/sipp/files/3775041/sipp.gz)). I've included an Alpine Linux Dockerfile that does such a build - publishing this binary as part of the release (and maybe pushing a SIPp container to Docker Hub?) will hopefully cut down the number of people who need to build from source and therefore cut down the number of issues raised about it.

This also moves CI from Travis to Github Actions.

Still to do:

- [x] sort out the USE_SCTP, USE_SSL and USE_EPOLL flags
- [x] get the UTs working
- [x] sort out Github Actions
- [x] test it on a Red Hat Linux
- [x] test it on Ubuntu Linux flavour
- [x] test it on FreeBSD
- [x] test it on OS X
- [x] test it on Windows
- [x] make checking out the gmock subdirectory/running the UTs optional


rkday:
@wdoekes - thoughts?

wdoekes:
At a quick glance, this looks good :+1: 

It's probably best if the old method of building stays working in 3.6.x (./configure, ./autogen, ./build), so we may want to fork/branch 3.6.x from an older commit if this breaks anything (and start 3.7.x). If it doesn't, we can do that branch just before/at #430.

(And I'll assume you'll be squashing this to fewer commits before merging.)

rkday:
Done - https://github.com/SIPp/sipp/tree/branch/3.6.

PR #430, opened 2019-10-27
Delete autoconf/automake tooling

rkday:
Separate from #429, as we might want a period where both CMake and the old build tooling are available, to smooth over transition problems. (But maybe if someone can't build with CMake they should just use an old release and raise an issue.)


wdoekes:
Like I said in #429: let us branch off 3.6.x and make master 3.7.x-dev before merging this.

Issue #413, opened 2019-08-09
Terminal output is unreadable in MacOS (Catalina)


stemis:
Hello,

I have Mac Version: `10.15 (19A512f)`
SIPp version: `SIPp v3.6.0-PCAP-RTPSTREAM`

When I run `sipp -sn uac <IP>`, the data printed to the terminal is not handled correctly and it is constantly printing new lines, making the output completely unreadable.

Please see an example screenshot below

![Sipp](https://user-images.githubusercontent.com/4124979/62781885-3d335080-bab9-11e9-8788-10eea18970c9.png)

The screen constantly jumps making it impossible to read anything

wdoekes:
@stemis: I'm sorry, I've already tried to fix problems in this area once before (see #76 and #102). But as I have no OSX, I don't have any means to reproduce.

rkday:
@stemis - `NCURSES_NO_SETBUF=1 ./sipp -sn uas` seems to stabilise it a bit for me on my Mac. I still see the symptoms from https://github.com/SIPp/sipp/pull/102#issuecomment-57734282, ocasional "rolling", but it's a lot more usable.

rkday:
I have the beginnings of what seems to be a fix in https://github.com/SIPp/sipp/pull/431.

rkday:
@stemis - if you can pull and build the latest code from https://github.com/SIPp/sipp/pull/431 and see if that works for you, that would be useful. It works on my Mac, but more testing is always good!

rkday:
Fixed by #431 .

PR #431, opened 2019-10-29
Use ncurses in a more standard and centralised fashion

rkday:
This changes how we draw to screen - instead of a bunch of `printf` statements, we render things line-by-line into a `std::vector<std::string>`. Once we have that, we can print it to screen, write it to a file, etc. as we please.

This should make problems like https://github.com/SIPp/sipp/issues/413 easier to debug - the number of lines of code printing stuff to the screen will be much smaller, so bugs will have fewer places to hide, and it'll be easier to inspect what's being printed by looking at the vector. (And indeed that issue doesn't repro on this branch.)

It cleans up the responsibilities of a few classes/files:

- a new ScreenPrinter class, in screen.cpp, is responsible for the curses UI
- CStat/stat.cpp loses responsibility for printing the stat and repartition screens, and just tracks/calculates stats
- logger.cpp loses its curses UI code but gains the error logging code that was in screen.cpp 

I've also moved to more standard curses APIs - `printw` instead of `printf`, `clear()` instead of `printf("\033[2J")`, etc.

Todo:

- [x] Complete changes for the scenario screen
- [x] Move the statistics screen to the new model
- [x] Move the repartition screen to the new model
- [x] Move the variables screen to the new model
- [x] Move the TDM map screen to the new model
- [x] Move the secondary repartition screens to the new model
- [x] Get -trace_screen working in this mode
- [x] Print out a final summary to stdout _after_ disabling curses


rkday:
@wdoekes, do you think I should go ahead and merge this or do you want me to hold off to give you time to review? (It's pretty big but mostly moving stuff around.)

I've tested:

- on Linux and OS X
- by tabbing through the various screens
- with -trace_screen
- with stdout not being a TTY (courtesy of the regression tests)
- under valgrind to check there are no crass memory bugs in the mainline case

wdoekes:
Quick review says: LGTM :)

src/sipp.cpp
```diff
@@ -990,7 +990,9 @@ static void help_stats()
 static void print_last_stats()
 {
     interrupt = 1;
-    sp->print_closing_stats();
+    if (sp) {
+      sp->print_closing_stats();
```
wdoekes: I see the indentation check doesn't work flawlessly ;)

wdoekes:
```
diff --git a/validate-src.sh b/validate-src.sh
index f720ffb..80ba0cf 100755
--- a/validate-src.sh
+++ b/validate-src.sh
@@ -49,13 +49,13 @@ fi
 files_with_not_4_spaces=$(
     find include src '!' -name config.h \
       '(' -name '*.cpp' -o -name '*.hpp' -o -name '*.c' -o -name '*.h' ')' |
-    xargs grep -lE '^(  \}| {1,3}[a-z_])')
+    xargs grep -lE '^(    )* {1,3}[^ *]')
 if test -n "$files_with_not_4_spaces"; then
     echo "files with non-standard indentation found:" >&2
     for file in $files_with_not_4_spaces; do
         echo "  $file" >&2
         if $print_listing; then
-            grep -nE '^(  \}| {1,3}[a-z_])' $file |
+            grep -nE '^(    )* {1,3}[^ *]' $file |
                 sed -e "$list_sed" | cat -A >&2
         fi
     done
```

rkday: I've actually been thinking of switching to something more hardened like clang-format (once the count of open pull requests goes down a bit, because it'll cause merge conflicts).

PR #445, opened 2019-12-05
Should not access SDP body when 'Content-Length: 0'

sunlin7:
The call::process_incoming() emit error message when 'Content-Type: application/sdp' is presented but 'Content-Length: 0'. Shouldn't try to access the SDP body for this situation.

### DIFF
--- a/src/call.cpp
+++ b/src/call.cpp
@@ -2822,7 +2822,10 @@ bool call::process_incoming(const char* msg, const struct sockaddr_storage* src)
     /* Check if message has a SDP in it; and extract media information. */
     if (!strcmp(get_header_content(msg, "Content-Type:"), "application/sdp") &&
             hasMedia == 1 && !curmsg->ignoresdp) {
-        extract_rtp_remote_addr(msg);
+        const char* ptr = get_header_content(msg, "Content-Length:");
+        if (ptr && atoll(ptr) > 0) {
+            extract_rtp_remote_addr(msg);
+        }
     }
 #endif
 


wdoekes:
Thanks for the PR.

What is the error that is emitted?

And, note that you cannot use xp_get_content_length for this purpose. It might do what you want (now), but it's for parsing the scenario file, not for parsing incoming SIP.

You'll want to use `get_header_content(msg, "content-length")` instead (which returns a string, so you might want to wrap that so you can do integer comparison).

sunlin7:
Hi @wdoekes , The error message is from the next call call::extract_rtp_remote_addr() and message is "extract_rtp_remote_addr: invalid IP version in SDP message body".
I'll change the to get_header_content() the update the PR.

src/call.cpp
```diff
@@ -2821,7 +2821,9 @@ bool call::process_incoming(const char* msg, const struct sockaddr_storage* src)
 #ifdef RTP_STREAM
     /* Check if message has a SDP in it; and extract media information. */
     if (!strcmp(get_header_content(msg, "Content-Type:"), "application/sdp") &&
-            hasMedia == 1 && !curmsg->ignoresdp) {
+        hasMedia == 1 && !curmsg->ignoresdp) {
+      const char *ptr = get_header_content(msg, "content-length:");
```
wdoekes: Please use "Content-Length" instead of "content-length", like the other users of get_header_content do.

And use the C++-style pointer placement (`char* ptr` instead of  `char *ptr`).

src/call.cpp
```diff
@@ -2821,7 +2821,9 @@ bool call::process_incoming(const char* msg, const struct sockaddr_storage* src)
...
+      if (ptr && atoll(ptr) > 0)
```
wdoekes: Please add curly braces, even for single line ifs, like the https://github.com/SIPp/sipp/#contributing readme says.

src/call.cpp
```diff
@@ -2821,7 +2821,9 @@ bool call::process_incoming(const char* msg, const struct sockaddr_storage* src)
...
         extract_rtp_remote_addr(msg);
```
wdoekes: Please reindent everything with 4 spaces (see also readme).

sunlin7:
Hi @wdoekes , the modification has chagned as your wish, please review again. Thank you.

wdoekes:
Thanks for the contribution!

PR #450, opened 2020-02-07
RTPStream: Skip wav header when playing

orgads:


wdoekes:
While I'm not against making things more userfriendly, I'm not really sure when this helps. (And rather, wonder if it might hurt.)

- Who plays WAV files when you already have to encode a file specifically for alaw/ulaw/gxxx with the right channel count, bits and frequency?
- Does the documentation (not) state that wav files (or raw) files are allowed?

Skipping a wav header here might be a sliding slope towards asserting that we "support" a particular format (optionally handling transcoding etc., which we don't).

orgads:
https://sipp.readthedocs.io/en/v3.6.0/scenarios/actions.html#media-rtp-commands

Most of the examples you'll find on Google also refer to wav files.

The docs state (more or less) that the format should match the negotiated coder in the SDP.

wdoekes:
Hm.. I'm not convinced. Those docs also say "The audio file should be the raw samples, example files are included for PCMA, G722 and iLBC (mode=30)."

Can't we fix the docs to not say "wav"? (Or even, loudly complain if the file starts with a RIFF header.)

I fear implementing partial wav support only opens up a can of worms.

orgads:
I see your point, but I have to say that I do find it convenient to be able to play the file with any media player. The restrictions are clear, and I don't think anyone expects sipp to support transcoding.

wdoekes:
@orgads:
```
$ hd /home/walter/src/cpython/Lib/test/audiodata/pluck-pcm32.wav | head
00000000  52 49 46 46 de 67 00 00  57 41 56 45 66 6d 74 20  |RIFF.g..WAVEfmt |
00000010  10 00 00 00 01 00 02 00  11 2b 00 00 88 58 01 00  |.........+...X..|
00000020  08 00 20 00 4c 49 53 54  5a 00 00 00 49 4e 46 4f  |.. .LISTZ...INFO|
00000030  49 4e 41 4d 06 00 00 00  50 6c 75 63 6b 00 49 41  |INAM....Pluck.IA|
00000040  52 54 12 00 00 00 53 65  72 68 69 79 20 53 74 6f  |RT....Serhiy Sto|
00000050  72 63 68 61 6b 61 00 00  49 43 4d 54 18 00 00 00  |rchaka..ICMT....|
00000060  41 75 64 61 63 69 74 79  20 50 6c 75 63 6b 20 2b  |Audacity Pluck +|
00000070  20 57 61 68 77 61 68 00  49 43 52 44 06 00 00 00  | Wahwah.ICRD....|
00000080  32 30 31 33 00 00 64 61  74 61 58 67 00 00 bc 65  |2013..dataXg...e|
00000090  2d 02 92 9d eb ff 80 0f  5a 4b 9c 54 fa 00 40 c3  |-.......ZK.T..@.|
```
That's the first WAV I checked, and I see it does not have a 44 bytes header.

However, parsing the chunks up to `data` seems straight forward:
```python
import struct
import sys
from32 = (lambda x: struct.unpack('<L', x)[0])
SEEK_CUR = 1
with open(sys.argv[1], 'rb') as fp:
    assert fp.read(4) == 'RIFF'
    file_len_minus_8 = from32(fp.read(4))
    assert fp.read(4) == 'WAVE'
    while True:
        chunk = fp.read(4)
        chunk_size = from32(fp.read(4))
        fp.seek(chunk_size, SEEK_CUR)
        print(chunk, chunk_size)
        if chunk == 'data':
            break
```
```console
$ python wav.py /home/walter/src/cpython/Lib/test/audiodata/pluck-pcm32.wav
('fmt ', 16)
('LIST', 90)
('data', 26456)
```
That would make the parsing at least a little bit less fragile and not choke on the first wav that has metadata.

wdoekes:
@orgads : care to make the desired changes to make your wav parsing less fragile?

orgads:
Oops, I totally forgot about it. I'll try to do it next week. Thanks for the heads up.

Issue #458, opened 2020-05-06
XML attribute values are not consistently unescaped (e.g. for <strcmp value>)


nud:
Most XML attributes are not unescaped by the XML parser. After checking the source code it looks like only `<ereg>` is.

While it has no practical consequences most of the time, it makes it impossible to e.g. use `<strcmp>` to do something like this:

```xml
  <strcmp
      assign_to="result"
      variable="strvar"
      value="&#34;Some User&#34; &lt;sip:foo@bar.com&gt;"
      />
```
I suppose solving https://github.com/SIPp/sipp/issues/31 would fix the problem globally, but in the meantime unescaping at least attributes that are meant to contain strings would be a nice improvement.

wdoekes:
Thanks @nud !

PR #459, opened 2020-05-06
Consistently unescape XML attributes when loading scenario

nud:
This changes the behaviour of `xp_get_string()` in order to
systematically unescape the returned values, as other XML parsers do.
The custom unescaping in the handling of the `<ereg>` element has been
removed as it is now unneeded.

This will make it possible to e.g. compare strings that contain special
XML characters using `<strcmp>`.

A regression test case has been added to highlight the problem.

https://github.com/SIPp/sipp/issues/458

### DIFF
--- a/regress/github-#0458/run
+++ b/regress/github-#0458/run
@@ -1,0 +1,22 @@
+#!/bin/sh
+# This regression test is a part of SIPp.
+. "`dirname "$0"`/../functions"; init
+
+sippfg -m 1 -sf uas.xml -p 5070 >/dev/null 2>&1 &
+job=$!
+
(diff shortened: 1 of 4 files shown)


nud:
Apparently the test xml linting fails because the DTD is missing some elements...

wdoekes:
> A regression test case has been added to highlight the problem.

:drooling_face: 

Awesome.

I can look into that lint-xml thingy.

nud:
It is worth noting that this PR does not fix escaping of number-encoded entities (i.e. escaping `"` with `&quot;` works, but escaping it with `&#34;` does not).

PR #503, opened 2021-01-06
Fix a segmentation fault in 'call.cpp:2369: createAuthHeader'

MGoetzl:
The SegFault only happens if the SIPp reference to the authentication header (in our case [field2], see the attached 'fail.xml') is used (in the example 'fail.xml': echo [field2]) after or in the receive section of the REGISTER message which actually sends the authentication header. In this case "getMethod" returns 0x0 (see 'gdb.txt', also attached).
It was introduced by commit 6dd80fe by using the function 'src->getMethod()' instead of the variable 'method'.

[fix_info.tar.gz](https://github.com/SIPp/sipp/files/5775910/fix_info.tar.gz)

### DIFF
--- a/src/auth.cpp
+++ b/src/auth.cpp
@@ -273,8 +273,10 @@ static int createAuthResponseMD5(
 
     // Load in A2
     md5_init(&Md5Ctx);
-    md5_append(&Md5Ctx, (md5_byte_t *) method, strlen(method));
-    md5_append(&Md5Ctx, (md5_byte_t *) ":", 1);
+    if (method) {
+        md5_append(&Md5Ctx, (md5_byte_t *) method, strlen(method));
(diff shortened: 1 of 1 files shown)


wdoekes:
So it crashes on `<exec command="echo [field2]"/>` ?

In that case the fix is not to calculate a different (wrong) hash, but do something else (return error?).

MGoetzl:
> 
> 
> So it crashes on `<exec command="echo [field2]"/>` ?

yes, it crashes because it will call "createAuthHeaderMD5" (auth.cpp:242) with "method" being a null pointer,  whitch leads to the crash in "strlen(method)" (auth.cpp:281)

> 
> In that case the fix is not to calculate a different (wrong) hash, but do something else (return error?).

Ah, yes - true! I've updated the pull request accordingly. It returns an error in the calling function "createAuthHeader" (auth.cpp:145).

Issue #32, opened 2013-10-07
OpenSSL should be replaced by GnuTLS


rkday:
OpenSSL is not GPL-licensed, which causes problems for SIPp distributors. We should use GnuTLS (http://www.gnutls.org/) for handling TLS mode instead.

linuxmaniac:
After 4 years, instead of porting to GnuTLS can you please add the license exception?
```
 * In addition, as a special exception, the copyright holders give
 * permission to link the code of portions of this program with the
 * OpenSSL library under certain conditions as described in each
 * individual source file, and distribute linked combinations
 * including the two.
 * You must obey the GNU General Public License in all respects
 * for all of the code used other than OpenSSL.  If you modify
 * file(s) with this exception, you may extend this exception to your
 * version of the file(s), but you are not obligated to do so.  If you
 * do not wish to do so, delete this exception statement from your
 * version.  If you delete this exception statement from all source
 * files in the program, then also delete it here.
```
https://people.gnome.org/~markmc/openssl-and-the-gpl.html

wdoekes:
Where do we add this?

wdoekes:
Something like that?

rkday:
@wdoekes , pretty sure you can't add that license exception - you're not the copyright holder for most of the code, so can't allow the code to be used in ways the original authors haven't granted permission for.

(In theory it would be possible to conact all the authors and get their permission, but that's a fairly substantial undertaking.)

wdoekes:
Hi Rob, thanks for pitching in.

You're probably right on that.

So. What are our options, apart from migrating to GnuTLS? Especially when we consider that we would be hard-pressed to find someone against this exception, right?

Would it be sufficient to contact the SIPp mailing list and request any original author to speak up if they are against? That, combined with an attempt to hunt down the most prolific authors of SIPp code could be enough to convince those with rights to agree that a decent attempt has been undertaken to respect everyones rights.

uhle:
Instead of porting the code to use GnuTLS, [WolfSSL](/wolfSSL/wolfssl) could be used with only few changes. Its OpenSSL portability layer is far more advanced and its license is GPL2 compliant. If you're interested, you could check out and try branch [wolfssl-support](/uhle/sipp/tree/wolfssl-support) or have a look at sipp@851d1fdbca795ebea5c646512695a723789f7d19.

I could also make a pull request if anyone wants me to.

wdoekes:
Yes please!

[...]

PR #505, opened 2021-01-14
Add support for WolfSSL as a GPLv2 compliant alternative to OpenSSL

uhle:
[WolfSSL](/wolfSSL/wolfssl) is an SSL/TLS library that comes with the WolfCrypt library and with a quite advanced OpenSSL compatibility layer which provides enough features to compile sipp with SIPS and SRTP enabled. Though I have to admit that the old TLS protocol version 1.0 is not supported by WolfSSL unless explicitly compiled in which is definitely not the case for the pre-compiled packages shipped with Debian (and Ubuntu) based Linux distributions. That is why I have added guards to those instructions selecting TLS 1.0 and instead showing an error message that TLS 1.0 is not supported. Also setting the minimum protocol version to TLS 1.0 for the negotiation process has been changed to TLS 1.1 respectively (cf. sslsocket.cpp).

So maybe WolfSSL is therefore not a 100% replacement for OpenSSL but could be used as an alternative that is GPLv2 compliant. That is why I decided to simply prefer OpenSSL over WolfSSL in CMakeLists.txt (if both libraries are installed with development files) and not adding another configuration option. You can simply influence CMake's choice by either installing the development files for OpenSSL or those for WolfSSL.

All the other changes are more or less straight-forward, mainly adding a new macro `USE_WOLFSSL` similar to `USE_OPENSSL` and adapting the code appropriately.

This pull request also addresses issue #32. I tried to use GnuTLS before but its OpenSSL compatibility layer is far behind WolfSSL's implementation and although I did manage to get quite a lot compiling and working, there are still some things that are not easily ported to GnuTLS (like support for CRL for instance). It would need so much more effort and I don't know if it's really worth it since WolfSSL is also actively maintained like OpenSSL or GnuTLS.


wdoekes:
Very very nice.

I'd say that most places that do a `defined(USE_OPENSSL) || defined(USE_WOLFSSL)` could use a `USE_TLS` instead. At least those where not open/wolfssl specific structures/types are used.

If you want to address that, then cool. Otherwise I'll probably step in and edit after merging.

uhle:
You're right. All these places in sipp.hpp/.cpp, message.hpp/.cpp, call.hpp/.cpp and rtpstream.hpp/.cpp are OpenSSL/WolfSSL agnostic and may be changed accordingly. I just wasn't quite sure if `USE_OPENSSL` had been used on purpose or just by chance. That's why I changed it to `defined(USE_OPENSSL) || defined(USE_WOLFSSL)`.

Do you want me to push yet another commit with these additional changes or do you prefer just a single commit, i.e. amend the previous commit and then force-push?

wdoekes:
> amend the previous commit and then force-push?

Yes please :)

PR #519, opened 2021-04-12
CMake: Bump minimum version to 2.8.12

orgads:
2.8 is deprecated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 2.8.12)
 
 # set the project name
 project(SIPp)


wdoekes:
Why would we want to force deprecation?

Are we using stuff that only works with 3.0+ ?

orgads:
There's a warning with cmake 3.20:

```
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 2.8.12 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

Can I change it to 2.8.12 then?

wdoekes:
Well. Either it's broken already, or it still works with 2.8.x.

- If it's already broken for 2.8.x, you'd be fixing a bug.
- If it's not broken for 2.8.x, you're removing the possibility for someone to install without any benefit (other than you not seeing the warning). (You'd be "creating a bug")

This change would only make sense to me if:

- you're adding to the CMakeLists.txt and (you suspect that you) add new features that will not be available in 2.8.x;
- or, you're revamping the CMakeLists.txt so it's 2.8.12+ compatible.

But you're doing neither.

If you're worried about versions of cmake newer than yours refusing to cmake, then perhaps the cmake_policy() is what you're after?
```patch
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 2.8)
+cmake_policy(VERSION 3.0)
 
 # set the project name
 project(SIPp)
```
As far as I understand the docs, that's the one that sets the compatibility mode, without forcing a user to upgrade their cmake when it's not necessary.

orgads:
Never mind. I'll live with this warning :)

PR #522, opened 2021-04-14
Stat: Dump buffered rtt data on end

orgads:
RTT data is buffered. Buffering can be controlled by rtt_freq, which
is 200 by default. The last calls that are buffered in a session are
never dumped, and this data was lost.


orgads:
I'm not sure if this is the right place. Feel free to suggest an alternative.

PR #523, opened 2021-04-21
Add URL encode/decode <action> for scenarios.

ticpu:
I really need this to decode the URI encoded Call-ID in the Replaces: header of a 3PCC scenario to send it via `<sendCmd>`.

### DIFF
--- a/docs/scenarios/actions.rst
+++ b/docs/scenarios/actions.rst
@@ -482,6 +482,24 @@ and microseconds since the epoch. For example::
 
 
 
+urlencode / urldecode
++++++++++++++++++++++
+
+The urlencode and urldecode actions will replace the content of the
+variable specified in variable with the coded version. For example::
+
+Content of variable_to_be_encoded is "this: is a string".
+
+    <nop>
+      <action>
+        <urlencode variable="variable_to_be_encoded" />
+      </action>
+    </nop>
+
+Content of variable_to_be_encoded now is "this%3A%20is%20a%20string".
+
+
+
 setdest
 +++++++
 
--- a/include/actions.hpp
+++ b/include/actions.hpp
@@ -66,6 +66,8 @@ public:
         E_AT_VAR_TO_DOUBLE,
         E_AT_VAR_STRCMP,
         E_AT_VAR_TRIM,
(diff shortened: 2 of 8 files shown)


docs/scenarios/actions.rst
```diff
@@ -482,6 +482,24 @@ and microseconds since the epoch. For example::
...
++++++++++++++++++++++
+
+The urlencode and urldecode actions will replace the content of the
+variable specified in assign_to with the coded version. For example::
+
+Content of variable2 is "this: is a string".
+
+    <nop>
+      <action>
+        <urlencode assign_to="variable2" />
```
wdoekes, line 495: This PR looks awfully complete :+1: 

I just wonder about this `assign_to`. Doesn't it make more sense to call it `variable`?
```xml
<urlencode variable="variable2" />
```

wdoekes (changes requested):

docs/scenarios/actions.rst
ticpu: Absolutely, I based my work on the trim method, that explains why I used assign_to :)
I'll give it a quick change.

docs/scenarios/actions.rst
```diff
@@ -482,6 +482,24 @@ and microseconds since the epoch. For example::
 
 
 
+urlencode / urldecode
++++++++++++++++++++++
+
+The urlencode and urldecode actions will replace the content of the
+variable specified in assign_to with the coded version. For example::
+
+Content of variable_to_be_encoded is "this: is a string".
```
wdoekes, line 491: Too fast ;)

docs/scenarios/actions.rst
```diff
@@ -482,6 +482,24 @@ and microseconds since the epoch. For example::
 
 
 
+urlencode / urldecode
++++++++++++++++++++++
+
+The urlencode and urldecode actions will replace the content of the
+variable specified in assign_to with the coded version. For example::
```
wdoekes, line 489: Too fast ;)

ticpu: Oops, too fast indeed, I didn't see your comment before requesting re-review. Fixed.

PR #550, opened 2021-10-30
Fix copy-paste issue in PR #549

uhle:
Replace `"audio"` by `"video"` where the second RTP stream creation is called to prevent misleading error messages.


wdoekes:
Thanks for the third pair of eyes :ok_hand:

Issue #559, opened 2021-11-15
[FreeBSD] make fails if -DUSE_PCAP was set on cmake


kadabusha:
Case on cmake parameter -DUSE_PCAP=1 was set, make fails on 40th. Same behavior is observed on 11.4-RELEASE-p9 and 11.3-RELEASE-p12 - if **DUSE_PCAP** is not specified, make finishes at no issues.
Checked on Debian10 - make finishes at no issues in same scenario.

```
[1@freebsd ~/sipp]$ pkg info|grep 'libpcap\|libnet\|curses\|cmake'
cmake-3.20.4                   Cross-platform Makefile generator
libnet-1.1.6_5,1               C library for creating IP packets
libpcap-1.10.0                 Ubiquitous network traffic capture library
ncurses-6.2.20210626           Library for terminal-independent, full-screen output
[1@freebsd ~/sipp]$ uname -r
11.4-RELEASE-p9
```

```
[1@freebsd ~]$ git clone https://github.com/SIPp/sipp.git
Cloning into 'sipp'...
remote: Enumerating objects: 7268, done.
remote: Counting objects: 100% (434/434), done.
remote: Compressing objects: 100% (270/270), done.
remote: Total 7268 (delta 283), reused 269 (delta 163), pack-reused 6834
Receiving objects: 100% (7268/7268), 5.71 MiB | 13.26 MiB/s, done.
Resolving deltas: 100% (5190/5190), done.
[1@freebsd ~]$ cd sipp
[1@freebsd ~/sipp]$ cmake . -DUSE_PCAP=1
-- The C compiler identification is Clang 10.0.0
-- The CXX compiler identification is Clang 10.0.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for endian.h
-- Looking for endian.h - not found
-- Looking for sys/endian.h
-- Looking for sys/endian.h - not found
-- Looking for sys/epoll.h
-- Looking for sys/epoll.h - not found
-- Performing Test HAVE_UDP_UH_PREFIX
-- Performing Test HAVE_UDP_UH_PREFIX - Failed
[...]

PR #562, opened 2021-11-22
CMakeLists.txt fix for installation on FreeBSD

kadabusha:
The change referenced to issue #559
Fix the installation on FreeBSD by usage of CMAKE_CXX_FLAGS instead of CMAKE_REQUIRED_FLAGS


rkday:
Looks plausible and doesn't break anything for me.

PR #569, opened 2022-02-08
Few additions to the code

smititelu:
Hello, we are using SIPp with the following features and I wonder if they are of interest for the upstream community:

1. Add define to use local ip hints

When USE_LOCAL_IP_HINTS is defined, prioritize local interface hints and
use them when deciding the remote address.

Useful when remote is a FQDN which can resolve to both IPv4 nd IPv6.

2. Generate random SSRC

3. Return ok if xml scenario finished ok. Log rtp or echo errors

4. Make TLS client cert/key file optional

Thank you,
Stefan


wdoekes:
Looks sane.

Could you split those up into separate PRs? Thanks!

smititelu:
Hi, yes will try splitting them in separate PRs.

smititelu:
Closing this PR as is, and will open new PRs for each task, and reference back this PR.

PR #570, opened 2022-02-28
Fix for unsupported TLS versions

svanstone-resilient:
Before this change, sipp would negotiate TLS 1.3. This doesn't seem to work though as sipp doesn't support TLS 1.3.
Now sipp will never negotiate a TLS version it doesn't support.


wdoekes:
Did you check where SIPp doesn't support TLS 1.3?

PR #574, opened 2022-03-19
Fix for local_port keyword using TCP or TLS

fsilvestre:
This is an old issue related in older versions and still without a definitive fix.
Thanks to Patrick Wakano who proposed it at Sipp-devel mailing list at 2013. [https://sourceforge.net/p/sipp/mailman/message/31219335/]

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -2566,6 +2566,11 @@ int open_connections()
         }
         sipp_customize_socket(tcp_multiplex);
 
+        /* This fixes local_port keyword value when transport are TCP|TLS and it's defined by user with "-p" */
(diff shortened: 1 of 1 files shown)


wdoekes:
Not entirely sure what this fixes. But it doesn't seem to break anything.

wdoekes:
Thanks!

atsushi-ageet:
I found that only on macOS, using the -t tn or -t ln options in SIPp always results in the following error:
```
2025-02-08 04:07:46.786826 1738955266.786826: Unable to bind TCP socket, errno = 48 (Address already in use)
```
After investigating the cause, it seems that this issue started occurring after this commit.
I confirmed that reverting this commit prevents the issue, but is this commit necessary?

PR #577, opened 2022-04-08
Feat/lua plugin

nfranzmeier:
This feature expands on the SIPP plugin capability so we can run lua scripts from inside SIPP and add/change/delete SIPP variables to be used the XML scripts, adds the ability to add new command line options, adds the capability to use rpc calls to get stats and fixes a memory leak.  This was all based on the 3.6 branch.

Here's a summary

1. Allow a user to create new "EXEC" commands which in this case are used to allow SIPP to interwork with a lua script to allow lua to add functionality to a running SIPP script to do things like lookup values in a DB, do RESTful API queries etc. which can change the values of SIPP variables and then send them back into the running script.  In this example the lua function is called my_function.

example:  <exec command="luathread@my_function calling [$5] called [$2] />
		<log  message="retkey [$retkey] calling = [$calling] called = [$called] />

!-- Do a loop looking for the data to be done -->

 <label id="start_of_loop"/>
    <pause milliseconds="100" />
    
    <nop>
             <action>
                <log message="luaread@[$retkey] 0" />
                <exec command="luaread@[$retkey] 0" />
                <strcmp assign_to="result" variable="msgDone" value="0" />
                <log message="result for msgDone  check = [$msgDone]"/>
             </action>
   </nop>

  
  <nop next="finished" test="result"/>

  
  <nop>
    <action>
      <add assign_to="counter" value="1"/>
      <test assign_to="flag_limit" variable="counter"
                compare="greater_than_equal" value="100"/>
        <log message="flag_limit=([$flag_limit])" />
    </action>
  </nop>

  <nop next="finished" test="flag_limit" />

  <nop next="start_of_loop"/>

 <label id="finished" />

    <nop>
        <action>
                
                <log message="finished!" />
                <log message="luaread@[$retkey] 1" />
                <exec command="luaread@[$retkey] 1" />
        </action>
  </nop>

When the luaread is complete, the values created in the lua script will be available to the running SIPP script for use in the XML.

2. In order to use a lua script the sipp command line, additional functionality can be added to the command line - also in the plugin.  This is done by adding a new dlsym function: "handle_args" which is passed the current SIPP command being processed along with it's argument.

Here is an example of how it will be called:  note the -plugin and -lua_file options as well as the -pid_file option.

sipp -plugin libmyapp.so -lua_file sipp_sa.lua -pid_file  /var/run/sipp_pg.pid -aa -ringbuffer_files 1000 -ringbuffer_size 50000000 -trace_msg -trace_logs -trace_err -message_file /var/log/sipp/sipp_message_sa.log -message_overwrite false -error_file /var/log/sipp/sipp_error_sa.log -error_overwrite false -calldebug_file /var/log/sipp/log/sipp_debug_sa.log -calldebug_overwrite false -log_file /var/log/sipp/sipp_sa.log -log_overwrite false -m 1000000 -l 1 -i 172.20.2.35 -p 5060 -s 8665988936 -nd -sf ./sa.xml 172.29.31.14

4. Note that the dlsym "init" has also been expanded to pass in the dlsym handle and the command line args and count so it can also process arguments.  The init dlsym command in our example is used to populate an internal table with various API function pointers which contains the new commands luarun - which will fork/exec a lua script, luathread which will run a lua script in a thread and then the results can be obtained using "luaread" as in the XML example above (by polling).  See the appDllLoad()  function in myapp.cpp.

5. Another dlsym command was added to allow overriding the automatic handling of messages like OPTIONS in case you want to add SIP headers etc. to those responses.

6. Added  the ability to do rpc calls so that various call stats can be retrieved by other programs or even from the command line using a rpc client program included in the examples directory - sa_ctl.
For example:  sa_ctl get_callcount will retrieve the call counts

7. Fixed memory leak associated with having AUTO RESPONSE on - leaked during OPTIONS messages


lemenkov:
I love the idea. Could you please rebase this PR on top of current master so we can proceed?

PR #584, opened 2022-09-06
[UAS] Copy Record-Route from INVITE to 200OK to comply with rfc3261

kadabusha:
When a UAS responds to a request with a response that establishes a
   dialog (such as a 2xx to INVITE), the UAS MUST copy all Record-Route
   header field values from the request into the response (including the
   URIs, URI parameters, and any Record-Route header field parameters,
   whether they are known or unknown to the UAS) and MUST maintain the
   order of those values.


wdoekes:
Thanks!

PR #591, opened 2022-10-17
Add 'transport' to the Contact header for UAC scenarios

martin-flaska:
Adding missing `;transport=[transport]` into Contact header for all UAC scenarion.
Not sure which all files I have to change, but my idea was to use also TCP transport with SIPp UAC scenario.
The default one, generated with `sipp -sd uac` is not working for me, because of this.


wdoekes:
Thanks!

Issue #586, opened 2022-09-11
next_url contains the display name


YannickLC:
SIPp release: SIPp v3.6.1-SCTP-PCAP-RTPSTREAM.

The Contact is:
`Contact: Bob <sip:bob@192.168.56.1:52166;ob>`

The [next_url] is:
`Bob <sip:bob@192.168.56.1:52166;ob>`

It should be:
`sip:bob@192.168.56.1:52166;ob`

If there is no display name, [next_url] is correctly formatted.

PR #594, opened 2022-11-16
Fix to [next_url] when a display name is present in the contact

enneig:
based on an earlier PR which seems to have regressed 
https://github.com/SIPp/sipp/pull/367

someone else opened an issue earlier:
https://github.com/SIPp/sipp/issues/586

I'm not familiar with this project tests so I skipped changing the test files. I have verified the fix with a local scenario.

Regards Rene


wdoekes:
Thanks!

PR #595, opened 2022-11-25
[aka-auth] handle binary passwords

szcom:
createAuthHeaderMD5 called from createAuthHeaderAKAv1MD5 may receive a password (RES) with 0s in it. using strlen() on this buffer results in wrong response calculated


wdoekes:
Looks sane. Thanks!

PR #596, opened 2022-11-29
print OpenSSL error reason when certificate load fails

rajeshsingh381:
When OpenSSL security level is not changed, this error may help to identify error reasons like short key size etc

### DIFF
--- a/src/sslsocket.cpp
+++ b/src/sslsocket.cpp
@@ -336,14 +336,14 @@ enum tls_init_status TLS_init_context(void)
     if (SSL_CTX_use_certificate_file(sip_trp_ssl_ctx,
                                      tls_cert_name,
                                      SSL_FILETYPE_PEM) != 1) {
-        ERROR("TLS_init_context: SSL_CTX_use_certificate_file failed");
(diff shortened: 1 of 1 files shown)


wdoekes (changes requested):
Please use thread safe versions in this multithreaded application. Thanks!

ERR_get_error is fine. ERR_error_string is not.

Even though this bit is not used simultaneously, we don't want others to copy non-thread-safe code to elsewhere later on.

src/sslsocket.cpp
```diff
@@ -336,14 +336,18 @@ enum tls_init_status TLS_init_context(void)
     if (SSL_CTX_use_certificate_file(sip_trp_ssl_ctx,
                                      tls_cert_name,
                                      SSL_FILETYPE_PEM) != 1) {
-        ERROR("TLS_init_context: SSL_CTX_use_certificate_file failed");
+        char errbuf[256] = { '\0' };
+        ERR_error_string_n( ERR_get_error(), errbuf, sizeof(errbuf) );
```
wdoekes, line 340: Please don't add extra spaces. `ERR_error_string_n(ERR_get_error(), errbuf, sizeof(errbuf));` is perfectly readable, as is `{'\0'};` :)

wdoekes (changes requested):

wdoekes:
Thanks!

PR #598, opened 2022-12-06
Add define to use local ip hints

smititelu:
When USE_LOCAL_IP_HINTS is defined, prioritize local interface hints and use them when deciding the remote address.

Useful when remote is a FQDN which can resolve to both IPv4 nd IPv6.

related to #569


lemenkov:
LGTM. Let's merge it.

PR #600, opened 2022-12-06
Add a log when sipp scenario ends due to rtp or echo errors

smititelu:
Previously, when got some rtp/echo errors, sipp returned a failure code upon successful completion of xml scenario.

Now it returns success if xml scenario completes, and just logs rtp errors.

related to #569


wdoekes (commented):
Not fond of this.

I would like to see both the (xml) scenario and the RTP status in the exit code.

If the RTPCHECK_FAILED is a bit that is OR'ed (e.g. `64`), you could ignore that and get the XML status (e.g. `(status & ~64) == 0`).

smititelu:
Let's see if I understood correctly:

If an SIPp xml scenario completes successfully from SIP point of view, but it had rtp errors, it should return (EXIT_TEST_OK|EXIT_RTPCHECK_FAILED)...

... and then, on OS side, check the return code of last ran (sipp) command and take out the EXIT_RTPCHECK_FAILED (e.g. $? & ~EXIT_RTPCHECK_FAILED)  ?

If that is the case, I don't agree with it. Because and end user will see sipp scenario complete ok but "$?" OS var will be != 0. So basically, sipp scenario failed, even if you see it completing successfully. That would be misleading, imho.

smititelu:
Updated PR to just log when scenario ends due to rtp/echo errors.

src/sipp.cpp
```diff
@@ -1134,6 +1134,7 @@ void sipp_exit(int rc, int rtp_errors, int echo_errors)
         // then everything is OK!
         if ((rtp_errors > 0) || (echo_errors > 0))
         {
+            ERROR("GOT rtp_errors = %d and echo_errors = %d", rtp_errors, echo_errors);
```
wdoekes, line 1137: ERROR() does not return, so EXIT_RTPCHECK_FAILED is not returned to the OS. You'd want WARNING.

Maybe `"Got %d rtp_errors and %d echo_errors"` for more pleasant english.

smititelu:
Updated log to warning

PR #601, opened 2022-12-06
Make TLS client cert/key file optional

smititelu:
Allow start of sipp TLS client, without a mandatory cert/key.

related to #569


src/sslsocket.cpp
```diff
@@ -131,6 +131,21 @@ SSL* SSL_new_client()
...
+    if (SSL_CTX_use_certificate_file(sip_trp_ssl_ctx,
+                                     tls_cert_name,
+                                     SSL_FILETYPE_PEM) != 1) {
+        ERROR("SSL_new_server: SSL_CTX_use_certificate_file failed");
+        return NULL;
+    }
+
+    if (SSL_CTX_use_PrivateKey_file(sip_trp_ssl_ctx,
+                                     tls_key_name,
+                                     SSL_FILETYPE_PEM) != 1) {
```
wdoekes, line 144: For indentation all the way to the right, please indent exactly to after the `(`, like above.

src/sslsocket.cpp
```diff
@@ -332,38 +347,6 @@ enum tls_init_status TLS_init_context(void)
...
-    }
-
-    if (SSL_CTX_use_certificate_file(sip_trp_ssl_ctx_client,
-                                     tls_cert_name,
-                                     SSL_FILETYPE_PEM) != 1) {
-        char errbuf[256] = {'\0'};
-        ERR_error_string_n(ERR_get_error(), errbuf, sizeof(errbuf));
-        ERROR("TLS_init_context: SSL_CTX_use_certificate_file (client) failed: %s", errbuf);
-        return TLS_INIT_ERROR;
-    }
```
wdoekes, line 352: Is a client cert now optional? Or is it not possible anymore?

smititelu:
Updated PR to use client cert only if key/cert files exist at that path.

PR #259, opened 2016-10-01
Snom patch: rtp_echo action

pbertera:
**rtp_echo:** This patch introduces a new action <rtp_echo> allowing to stop and start the RTP echo from the scenario file.

**Example:**
```xml
<nop>
        <action>
                <rtp_echo value="0"/> <!-- stop the RTP echo -->
        </action>
</nop>
<pause milliseconds="15000"/>
<nop>
        <action>
                <rtp_echo value="1"/> <!-- start the RTP echo -->
        </action>
</nop>
```
*All the attribution for this patch should go to Snom Technology AG*

PR #196, opened 2016-03-08
Remove lots of unnecessary RTPstream code and use [media_port] instead.

wdoekes:
If you use multiple streams on a UAS which also uses a single port: then the RTP would all have the same 4-tuple and be a giant stream mess. But, if you can live with that, this cleans up the code quite nicely.

Closes #192.

PR #607, opened 2023-02-20
Resolve full path of pcap which starts with '~'

rajeshsingh381:
expand paths like '~/path/to/file.pcap' to '/home/myuser/path/to/file.pcap'


rajeshsingh381:
Really wish this feature gets added in next release. To that end, any suggestions/comments to code implementation are welcome, I will try my best to cater them,

rajeshsingh381:
Hi @wdoekes , could you please take a look at this PR when you find some time and offer your suggestions, if any? I am new to github community, so please also do let me know if a direct ping is inappropriate, I will refrain from in future.

wdoekes:
Maybe add expanduser(). That way `~someuser/blah` would also work, which will be expected.

Chatgpt was helpful enough to give me this ("can you show me a C implementation of python os.path.expanduser?"):
```c
#include <stdlib.h>
#include <unistd.h>
#include <sys/types.h>
#include <pwd.h>

char* expanduser(char* path) {
    if (path == NULL) {
        return NULL;
    }
    
    if (path[0] == '~') {
        char* homedir = NULL;
        
        if (path[1] == '\0' || path[1] == '/') {
            homedir = getenv("HOME");
        } else {
            struct passwd* pw = getpwnam(path+1);
            
            if (pw != NULL) {
                homedir = pw->pw_dir;
            }
        }
        
        if (homedir != NULL) {
            char* expanded = malloc(strlen(homedir) + strlen(path));
            sprintf(expanded, "%s%s", homedir, path+1);
            return expanded;
        }
    }
    
    return strdup(path);
}
```
with these caveats:
> Note that this implementation uses malloc() and free() to allocate and deallocate memory for the expanded path, so you'll need to remember to free the returned string when you're done with it.

(I'd probably prefer a version with a provided buffer.)

(Also, it uses a non-reentrant `getpwnam`. Look into `getpwnam_r` instead..)

wdoekes:
```c
            sprintf(expanded, "%s%s", homedir, path+1);
```
Hehe. This can't possibly work for `~someuser` unless the username added too: `/home/someusersomeuser/bla`

(Also.. no malloc() NULL check. But I don't want any malloc here anyway.)

rajeshsingh381:
Thankyou @wdoekes for your suggestions, I have added support for `~/someuser/blah`, `~someuser/blah` and `~someuser` cases and tested them. Please let me know if you have any suggestions/inputs.

src/actions.cpp
```diff
@@ -531,6 +602,19 @@ void CAction::setPcapArgs (pcap_pkts  *  P_value)
...
+        if (expand_user_path(P_value, pcap_absolute_path))
+        {
+            P_value = strndup(pcap_absolute_path, strlen(pcap_absolute_path));
+        }
+        else
+        {
+            ERROR("Couldn't resolve pcap file absolute path: %s", P_value);
+        }
+    }
+
```
wdoekes, line 617: This would make a lot more sense to me:
```diff
--- a/src/send_packets.c
+++ b/src/send_packets.c
@@ -95,13 +95,20 @@ float2timer(float time, struct timeval *tvp)
 
 static char* find_file(const char* filename)
 {
+    char tmppath[MAX_PATH];
     char *fullpath;
-    if (filename[0] == '/' || !*scenario_path) {
-        return strdup(filename);
+    size_t len;
+
+    tmppath[0] = '\0';
+    expand_user_path(filename, tmppath, sizeof(tmppath));
+
+    if (tmppath[0] == '/' || !*scenario_path) {
+        return strdup(tmppath);
     }
 
-    fullpath = malloc(MAX_PATH);
-    snprintf(fullpath, MAX_PATH, "%s%s", scenario_path, filename);
+    len = strlen(scenario_path) + strlen(tmppath) + 1;
+    fullpath = malloc(len);
+    snprintf(fullpath, len, "%s%s", scenario_path, tmppath);
 
     if (access(fullpath, R_OK) < 0) {
         free(fullpath);
```
If you start from there, I think the rest will fall into place.

wdoekes (changes requested):
Please check my suggestion for the expand_user_path invocation. It should not be optional based on ~, but always be called directly.

wdoekes:
FYI: I moved find_file() to [fileutil.c](https://github.com/SIPp/sipp/blob/e86bf93947a50646882f7cc63f2680dd904de675/src/fileutil.c) where is used by both pcap and rtpstream file lookups.

rajeshsingh381:
Sorry for inactivity, I have applied the changes you suggested, and will verify the them over the different cases you mentioned tomorrow

rajeshsingh381:
Hi, have verified the new changes, for the mentioned 3 cases. I tried to reproduce and debug the regression test failures of github-#0259 and github-#0196, but weirdly they don't happen when I run the SIPp commands manually, call is successful with 0 return code.

rajeshsingh381:
Hi, apologies for inactivity. I have fixed the regression test failures.

src/fileutil.c
```diff
@@ -102,22 +102,27 @@ char* find_file(const char* filename)
 {
     char tmppath[MAX_PATH];
     tmppath[0] = '\0';
+    const char *filepathptr = tmppath;
     expand_user_path(filename, tmppath, sizeof(tmppath));
+    if(tmppath[0] == '\0')
+    {
+        filepathptr = filename;
+    }
```
wdoekes, line 110: Please don't start doing enters before the opening curly:

https://github.com/SIPp/sipp?tab=readme-ov-file#contributing

And a space between `if (`.

And don't do:
```
{ // comment here
```
Please look at the surrounding code and format the style in a similar fashion.

wdoekes: In the entire PR please :)

rajeshsingh381: Thanks for reviewing. I will test all scenarios again (as I have made some fixes based on your previous comments that I didn't understand then) and remind you.

lemenkov:
LGTM and it passes the tests.

src/fileutil.c
rajeshsingh381: Hi @wdoekes I have finished testing. Please let me know if any more issues or room for improvement or merge it as you seem fit. Thanks!

wdoekes:
Thanks for the style fixes :+1: 

I'm unfortunately unable to review this timely. Unassigning. Leaving the review+merge to someone else. E.g. @lemenkov

wdoekes (approved):

src/fileutil.c
```diff
@@ -15,29 +15,94 @@
...
+            home_dir = getenv("USERPROFILE");
+        } else {
+            snprintf(expanded_home_path, buflen - 1, "%s%s", home_dir, path + 1);
+        }
+    } else {
+        const char* first_slash = NULL;
+        first_slash = strchr(path, '/');                                                        /* substring starting from '/' */
+        size_t linux_username_limit = 32;                                                       /* As of now */
+        char* username = NULL;
+        if ((first_slash != NULL) && ((first_slash - (path + 1)) <= linux_username_limit)) {    /* '~someuser/blah' case */
```
orgads, line 48: I think you can remove `!= NULL` and the parentheses around `first_slash`.

src/fileutil.c
```diff
@@ -15,29 +15,94 @@
...
+            username = strndup(path + 1, first_slash - (path + 1));
+        } else {                                                                                /* '~someuser' case, there is no file, just username */
+            return -1;
+        }
+
+        struct passwd pwd;
+        struct passwd* result;
+        size_t bufsize = sysconf(_SC_GETPW_R_SIZE_MAX);
+        char buffer[bufsize + 1];
+        int retcode = getpwnam_r(username, &pwd, buffer, bufsize, &result);
```
orgads, line 58: You can `free(username)` right after this line, right?

src/fileutil.c
```diff
@@ -15,29 +15,94 @@
...
+        if ((first_slash != NULL) && ((first_slash - (path + 1)) <= linux_username_limit)) {    /* '~someuser/blah' case */
+            username = strndup(path + 1, first_slash - (path + 1));
+        } else {                                                                                /* '~someuser' case, there is no file, just username */
+            return -1;
+        }
+
+        struct passwd pwd;
+        struct passwd* result;
+        size_t bufsize = sysconf(_SC_GETPW_R_SIZE_MAX);
+        char buffer[bufsize + 1];
```
orgads, line 57: Dynamic array size is a GNU extension. Do we have this elsewhere in the codebase?

src/fileutil.c
```diff
@@ -15,29 +15,94 @@
...
+        char* username = NULL;
+        if ((first_slash != NULL) && ((first_slash - (path + 1)) <= linux_username_limit)) {    /* '~someuser/blah' case */
+            username = strndup(path + 1, first_slash - (path + 1));
+        } else {                                                                                /* '~someuser' case, there is no file, just username */
+            return -1;
+        }
+
+        struct passwd pwd;
+        struct passwd* result;
+        size_t bufsize = sysconf(_SC_GETPW_R_SIZE_MAX);
```
orgads, line 56: const

src/rtpstream.cpp
```diff
@@ -1814,7 +1814,7 @@ static int rtpstream_get_localport(int* rtpsocket, int* rtcpsocket)
     int port_number;
     int tries;
     struct sockaddr_storage address;
-    int max_tries = (min_rtp_port < (max_rtp_port - 2)) ? 100 : 1;
+    int max_tries = (min_rtp_port < (max_rtp_port - 2)) ? (max_rtp_port - min_rtp_port) : 1;
```
orgads, line 1817: This is unrelated

orgads (changes requested):

src/fileutil.c
rajeshsingh381: Thanks for suggestion. But I believe it will be optimized by compiler anyway and in my opinion `if (first_slash != NULL)`, or even better `if (NULL != first_slash)`, is more readable than `if (first_slash)`. What do you think? However, If it is a coding guideline for this repo, I will oblige and make changes as you suggest.

src/fileutil.c
rajeshsingh381: Done, Thanks! will push all changes at once

src/fileutil.c
orgads: I don't mind. If you find it more readable, that's fine with me.

src/fileutil.c
```diff
@@ -15,29 +15,94 @@
...
+    if (path[1] == '\0' || path[1] == '/') {                                                    /* '~/path' case */
+        home_dir = getenv("HOME");
+        if (home_dir == NULL) {
+            home_dir = getenv("USERPROFILE");
+        } else {
+            snprintf(expanded_home_path, buflen - 1, "%s%s", home_dir, path + 1);
+        }
+    } else {
+        const char* first_slash = NULL;
+        first_slash = strchr(path, '/');                                                        /* substring starting from '/' */
```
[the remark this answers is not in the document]
rajeshsingh381: I googled a bit but couldn't find any reason not to, could you please tell me if there is any disadvantage of explicitly initializing it to `NULL` , so that I won't forget it in future.

src/fileutil.c
rajeshsingh381: I was not aware of this and couldn't find such examples in the codebase either. Now using `malloc()` reluctantly.

src/fileutil.c
rajeshsingh381: Done, Thanks!

src/rtpstream.cpp
rajeshsingh381: This was added during a upstream merge, not sure why was it not removed when I merged upstream again. Reverted in my repo now, hopefully the PR commits will be squashed.

rajeshsingh381:
Hi @orgads could you please review the updated diff when you get some time. Thankyou.

src/fileutil.c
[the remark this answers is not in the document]
orgads: But if you assign it directly, it *is* initialized. I meant: 
```c
const char* first_slash = strchr(path, '/');                                                        /* substring starting from '/' */
```

orgads (approved):
Thanks for the reminder. 

I approve, just reply on the NULL comment and then I'll merge.

src/fileutil.c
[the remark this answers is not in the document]
rajeshsingh381: Sorry for not catching it earlier, updated now. Thanks!

orgads:
You have conflicts. Please rebase.

PR #613, opened 2023-03-10
Fix crash

GangZhuo:
When reusing the RTP thread, a thread with ID=0 may be added to the threadIDs.
This causes to pthread_join() on 0 when shutdown RTP stream,
leading to the program crash and failure to write screen logs, etc.

### DIFF
--- a/src/rtpstream.cpp
+++ b/src/rtpstream.cpp
@@ -1807,7 +1807,9 @@ static int rtpstream_get_localport(int* rtpsocket, int* rtcpsocket)
 
     debugprint("rtpstream_get_localport\n");
 
-    next_rtp_port = min_rtp_port;
+    if (!next_rtp_port) {
+        next_rtp_port = min_rtp_port;
+    }
 
     /* initialise address family and IP address for media socket */
     memset(&address, 0, sizeof(address));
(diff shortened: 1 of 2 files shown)


wdoekes:
- Apparently, removeRtpTaskThreadID() is not used at all in the codebase (could be removed).
- Looks like threadID should not be 0:
  ```
  b2f7d2ac src/call.cpp (Jeannot Langlois      2020-08-09 16:02:09 -0400 6142)             // Obtain ID of parent thread used for the related RTP task
  b2f7d2ac src/call.cpp (Jeannot Langlois      2020-08-09 16:02:09 -0400 6143)             call_scenario->addRtpTaskThreadID(rtpstream_callinfo.threadID);
  ```
  I would think that we need something like this instead:
  ```diff
  --- a/src/rtpstream.cpp
  +++ b/src/rtpstream.cpp
  @@ -1461,6 +1461,8 @@ static int rtpstream_start_task(rtpstream_callinfo_t* callinfo)
           callinfo->threadID = threadID;
           /* Add thread to list of ready (spare capacity) threads */
           ready_threads[num_ready_threads++] = threaddata;
  +    } else {
  +        callinfo->threadID = the_selected_thread_id;
       }
   
       /* now add new task to a spare slot in our thread tasklist */
  ```

wdoekes:
@jeannotlanglois ^ maybe you have an opinion on this

rkday:
#621 implements @wdoekes' alternative fix instead, so I'm closing this. (Fine to reopen if that fix doesn't work.)

PR #614, opened 2023-03-13
Make static-builds work

### LINKED ISSUE #514: Github action build-static doesn't produce a static build
I've tried the build-static action from @rkday but the produced binary isn't static:

ldd sipp
	linux-vdso.so.1 (0x00007ffef1198000)
	libgsl.so.23 => not found
	libgslcblas.so.0 => not found
	libpcap.so.1 => not found
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fa20ac35000)
	libc.musl-x86_64.so.1 => not found
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fa20aae6000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fa20a8f2000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fa20b0fc000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fa20a8d7000)

p120ph37:
Resolves #514 

CMake 3.4 changes default behavior regarding static builds as part of https://cmake.org/cmake/help/latest/policy/CMP0065.html

Simply bumping the `cmake_minimum_required` value to `3.4` allows static-builds to work correctly.

Also note that Alpine has an incorrect `ldd` and `file` implementation which will still show the output executable as dynamic even though it is `static-pie`.  You can confirm the output is static by using `ldd` or `file` from Debian rather than from Alpine to inspect it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.4)
 
 # set the project name
 project(SIPp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,6 +47,11 @@ CHECK_SYMBOL_EXISTS(le16toh "sys/endian.h" HAVE_DECL_LE16TOH_BSD)
 configure_file("${PROJECT_SOURCE_DIR}/include/config.h.in"
                "${PROJECT_BINARY_DIR}/config.h" )
 
+if(BUILD_STATIC)
+  set(CMAKE_EXE_LINKER_FLAGS "-static-libgcc -static-libstdc++ -static")
+  set(CMAKE_FIND_LIBRARY_SUFFIXES .a ${CMAKE_FIND_LIBRARY_SUFFIXES})
+endif(BUILD_STATIC)
+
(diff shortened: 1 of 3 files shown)


p120ph37:
I also added a line to the `if(BUILD_STATIC)` block to put `.a` at the beginning of the list of lib suffixes to search for.  This corrects the indeterminate behavior of `find_library` where it will sometimes set `*_LIBRARY` to point to a dynamic (`*.so`) file that it happened to find first, even when there is a static library available.  When `BUILD_STATIC` is set, I add `.a` to the beginning of the existing `CMAKE_FIND_LIBRARY_SUFFIXES` list rather than replacing the list with just `.a`, since some platforms (e.g. Windows) may use a different library suffix, and I don't want to get in the way of making static-linking possibly work there in the future.

p120ph37:
Added some documentation on using the provided Dockerfile, and also provided an additional Dockerfile with the needed libraries and command for a static build with all features enabled.

rkday:
Thanks!

PR #617, opened 2023-03-31
fix verifyauth of passing empty auth when calling function verifyAuthHeader

liuzhaohui:
The`get_header` function is using a static variable `last_header`, which stores the parsed header or content. 
At every function call the variable `last_header` is first cleared and  then put the newly parsed header or content.

This will cause problem if we use `get_header` to return a data buffer, before really using it, another `get_header` call comes. 

In this case, before we call `verifyAuthHeader` which really use the data buffer, there is a call to `createSendingMessage` which will call `get_header_content` which finally call `get_header`.


rkday:
LGTM, thanks!

Issue #581, opened 2022-08-29
extract_rtp_remote_addr fails with an error when no rtp stream is found


hafkensite:
When no audio/video/image media line is found, the `extract_rtp_remote_addr` function calls ERROR. 
[extract_rtp_remote_addr](https://github.com/SIPp/sipp/blob/f44d0cf5dec0013eff8fd7b4da885d455aa82e0e/src/call.cpp#L333-L335)

The RFC does not seem to forbid that from happening, and in certain scenarios the PJSIP channel implementation from Asterisk can send
an INVITE without media when dealing with t38.

It is still a little odd, but should in my opinion not be an ERROR, but maybe a `WARNING`

Issue #565, opened 2021-12-03
rtpstream files not read during startup in 3.7


wdoekes:
3.6:
```
$ ~walter/src/sipp/sipp -m 1 -sf ~walter/srcelf/sipp-scenarios/out-in/iinv-o200-obye.xml -i 127.0.0.2
2021-12-03	09:02:02.723441	1638518522.723441: Cannot read/cache rtpstream file beep_1sec_50x160b.alaw
```
3.7:
```
$ ~walter/src/sipp/sipp -m 1 -sf ~walter/srcelf/sipp-scenarios/out-in/iinv-o200-obye.xml -i 127.0.0.2
... runs fine ... until it needs the alaw
```
Also: for pcaps we tried reading the scenario file directory instead of cwd. Maybe we should too for the rtpstream audio.

rkday:
I can't trivially reproduce the 'Cannot read/cache' behaviour on 3.6.1:

```
17:29:43 > $ ./sipp -v | head -n 2  

 SIPp v3.6.1-RTPSTREAM.
                                                                                                                                                                                                                   
17:29:46 > $ grep rtp_stream rtp.xml
    <exec rtp_stream=”file.wav” />
                                                                                                                                                                                                                   
17:29:48 > $ ls file.wav
ls: cannot access 'file.wav': No such file or directory
                                                                                                                                                                                                                   
17:30:29 > $ ./sipp -m 1 -sf rtp.xml 127.0.0.8
Resolving remote host '127.0.0.8'... Done.
------------------------------ Scenario Screen -------- [1-9]: Change Screen --
  Call rate (length)   Port   Total-time  Total-calls  Remote-host
  10.0(0 ms)/1.000s   5060       0.10 s            1  127.0.0.8:5060(UDP)

  Call limit 1 hit, 0.0 s period          0 ms scheduler resolution            
  0 calls (limit 30)                      Peak was 1 calls, after 0 s
  0 Running, 2 Paused, 0 Woken up
  0 dead call msg (discarded)             0 out-of-call msg (discarded)
  0 open sockets                          0/0/0 UDP errors (send/recv/cong)

                                 Messages  Retrans   Timeout   Unexpected-Msg
      INVITE ---------->         1         0         0                         
         100 <----------         0         0         0         1               
         180 <----------         0         0         0         0               
         183 <----------         0         0         0         0               
         200 <----------  E-RTD1 0         0         0         0               
         ACK ---------->         0         0                                   
[...]

Issue #564, opened 2021-12-02
Cannot play pcaps with sipp-3.7.0.rc1


Palezvar:
I just downloaded sipp-3.7.0.rc1.tar.gz and compiled it with --full option.

Trying to run a local test using the following two commands from the build directory:
1) ./sipp -sn uas
2) ./sipp docs/uac_pcap.xml -l 1 -m 1 -r 1 127.0.0.1

                                 Messages  Retrans   Timeout   Unexpected-Msg
0 :      INVITE ---------->         368       0         0
1 :         100 <----------         0         0         0         0
2 :         180 <----------         368       0         0         0
3 :         183 <----------         0         0         0         0
4 :         200 <----------  E-RTD1 368       0         0         0
5 :         ACK ---------->         368       0
6 :       Pause [      0ms]         368                           0
7 :         BYE ---------->         368       0         0
8 :         200 <----------         368       0         0         0

That default scenario has a couple of pauses and two pcaps to play but the output does not reflect that.

The signaling part is completed successfully but the UAC is not playing back the pcap g711a.pcap and dtmf_2833_1.pcap. I am trying to verify these via tcpdump. I am not seeing any RTP exchange. It does not follow any pause rules either.

am I missing anything?

Thanks,

Palezvar:
I just removed all the pcap related sections from the default uac_pcap.xml and only left a 8000 msec pause in there.

When I run the test locally: ./sipp docs/uac_pcap.xml 127.0.0.1

the pause stage read 0msec. I tried the same thing with the default uac.xml file and added a 8000msec pause and it again shows 0msec pause.

wdoekes:
You're not running `docs/uac_pcap.xml`; you have no `-sf`.

You're running `-sn uac`, which has a pause of 0ms:
```
$ ./sipp -sd uac | grep pause
  <pause/>
```

wdoekes:
I think you'd benefit from something like this:
```patch
--- a/src/sipp.cpp
+++ b/src/sipp.cpp
@@ -1377,10 +1377,16 @@ int main(int argc, char *argv[])
 #define CHECK_PASS() if (option->pass != pass) { break; }
 
     for (int pass = 0; pass <= 3; pass++) {
+       int have_remote_host = 0;
+
         for(argi = 1; argi < argc; argi++) {
             struct sipp_option *option = find_option(argv[argi]);
             if (!option) {
                 if (argv[argi][0] != '-') {
+                    if (have_remote_host) {
[...]

Issue #338, opened 2018-03-29
sipp stops working with error " Couldn't find 'Proxy-Authenticate' or 'WWW-Authenticate' in 401 or 407!"


tiglat:
Hello!

We met strange behaviour - after some time (depending on load from 15 min to several hours) sipp is suddenly terminated with error message " Couldn't find 'Proxy-Authenticate' or 'WWW-Authenticate' in 401 or 407!" However response we received contains such headers.

I tried to patch sipp source code and extended that log message with printing msg (code below). When the issue was reproduced msg was empty! 

        if (auth[0] == 0) {
            ERROR("Couldn't find 'Proxy-Authenticate' or 'WWW-Authenticate' in 401 or 407! - %s", msg);
        }

Please find my scenario in attachment.
[SendInviteAuthUAC.txt](https://github.com/SIPp/sipp/files/1860377/SendInviteAuthUAC.txt)

Similar scenario with REGISTER messages works fine.

wdoekes:
Sounds like a bug. Not enough resources to investigate further at this point though. Please do report the SIPp version that you encountered this problem with.

PR #620, opened 2023-04-01
Ensure that sockets are marked as non-blocking before OpenSSL calls are made

rkday:
A colleague reported an issue where:

- when using single-socket TLS (`-t l1`), SIPp hangs after printing the first screen
- that SIPp in this state will start working after resizing the terminal window that SIPp is running on
- and sending a SIGUSR2 has the same effect

The problem (which I diagnosed with `gdb`) was that SIPp was stuck in a SSL_read call:

```
(gdb) bt
#0  0x00007fbe2e21a392 in __libc_read (fd=5, buf=0x55ceafd526d3, nbytes=5) at ../sysdeps/unix/sysv/linux/read.c:26
#1  0x00007fbe2df4c3d9 in ?? () from /lib/x86_64-linux-gnu/libcrypto.so.1.1
#2  0x00007fbe2df4767e in ?? () from /lib/x86_64-linux-gnu/libcrypto.so.1.1
#3  0x00007fbe2df464d4 in ?? () from /lib/x86_64-linux-gnu/libcrypto.so.1.1
#4  0x00007fbe2df46aa7 in BIO_read () from /lib/x86_64-linux-gnu/libcrypto.so.1.1
#5  0x00007fbe2e198b91 in ?? () from /lib/x86_64-linux-gnu/libssl.so.1.1
#6  0x00007fbe2e19ce1e in ?? () from /lib/x86_64-linux-gnu/libssl.so.1.1
#7  0x00007fbe2e19a6d0 in ?? () from /lib/x86_64-linux-gnu/libssl.so.1.1
#8  0x00007fbe2e1a1c45 in ?? () from /lib/x86_64-linux-gnu/libssl.so.1.1
#9  0x00007fbe2e1aca3f in ?? () from /lib/x86_64-linux-gnu/libssl.so.1.1
#10 0x00007fbe2e1acb47 in SSL_read () from /lib/x86_64-linux-gnu/libssl.so.1.1
#11 0x000055ceada5d28a in SIPpSocket::empty (this=0x55ceafd4af50) at /home/developer/src/sipp/src/socket.cpp:874
#12 0x000055ceada62f33 in SIPpSocket::pollset_process (wait=1) at /home/developer/src/sipp/src/socket.cpp:2939
#13 0x000055ceada79636 in traffic_thread (rtp_errors=@0x7ffee793fea8: 0, echo_errors=@0x7ffee793feac: 0) at /home/developer/src/sipp/src/sipp.cpp:599
#14 0x000055ceada7e08a in main (argc=14, argv=0x7ffee7940388) at /home/developer/src/sipp/src/sipp.cpp:2147
``` 

Presumably this is because it’s never sent anything, so there’s nothing coming in to read, and it can’t send anything because it’s blocking on this read.

Resizing the window sends a SIGWINCH – this causes the SSL_read to be interrupted (and return EINTR), so SIPp can move on to the tasks that send messages, and then there’s a response so the subsequent SSL_read reads it, and everyone’s happy. Any other signal, e.g. SIGUSR2, would have the same effect.

This fix seems to work, marking the socket as nonblocking before we ever try and read from it.

I think this is needed for TLS and not TCP because, per https://stackoverflow.com/a/41948239, , “the SSL setup calls will examine the state of your fd and set private variables within the BIO and the SSL objects based on the blocking/non-blocking state of the fd”. A later bit of code sets O_NONBLOCK, and that’s good enough for TCP, but I think by that time OpenSSL has already put the socket in blocking mode, because it was in blocking mode at the time of the setup calls.

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -1263,6 +1263,9 @@ SIPpSocket::SIPpSocket(bool use_ipv6, int transport, int fd, int accepting):
     ss_ssl = NULL;
 
     if (transport == T_TLS) {
+        int flags = fcntl(fd, F_GETFL, 0);
+        fcntl(fd, F_SETFL, flags | O_NONBLOCK);
+
         if ((ss_bio = BIO_new_socket(fd, BIO_NOCLOSE)) == NULL) {
             ERROR("Unable to create BIO object:Problem with BIO_new_socket()");
         }


rkday:
@wdoekes , I'm thinking of tagging 3.7.0 once this, #618 and #619 are in - what do you think?

wdoekes:
I think there are some others that would be nice to have:

- https://github.com/SIPp/sipp/issues/564#issuecomment-985430371 - simple UI fix;
- #613 - seems crippling for the srtp code;
- #617 - unsure whether this is just a theoretical issue, but the suggested fix is easy enough.

Maybe 613 and 617 can go in as-is, but make a new ticket to fix them up properly.

wdoekes:
And #581 and maybe #565.

rkday:
OK:

- #564 is addressed by https://github.com/SIPp/sipp/pull/623
- #613 is addressed by https://github.com/SIPp/sipp/pull/621
- #617 actually looks fine to me - any concerns with me just merging it? (It's possible that this is the bug behind https://github.com/SIPp/sipp/issues/338).
- #581 is addressed by https://github.com/SIPp/sipp/pull/622
- #565 I couldn't repro the old behaviour (but https://github.com/SIPp/sipp/commit/d8899fca5beeb00dba0d7e40a622ddce0fb88f64 looks suspicious)

I'll probably merge all of #617 through #622 in the next day or two then tag 3.7.0 - shout if you'd like me to hold off, or if you get a repro for #565 and want that fix in too.

wdoekes:
Thanks!

I don't have time to work on these now. I think we can do without 565.

Go ahead and merge and tag :)

PR #622, opened 2023-04-01
Continue with a warning when SDP is missing m=audio and m=video

### LINKED ISSUE #581: extract_rtp_remote_addr fails with an error when no rtp stream is found

rkday:
Fixes #581.

### DIFF
--- a/src/call.cpp
+++ b/src/call.cpp
@@ -330,10 +330,6 @@ std::string call::extract_rtp_remote_addr(const char* msg, int &ip_ver, int &aud
         }
     }
 
-    if ((audio_port==0)&&(video_port==0)&&(image_port==0)) {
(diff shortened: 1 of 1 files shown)


wdoekes (commented):
lgtm

wdoekes (approved):

PR #630, opened 2023-04-15
Add `-bind_to_device` option

igankevich:
Hello!

I've added a new `-bind_to_device` option that binds the main socket to the specified network device. We needed this option in a benchmark where each SIP client has to have a unique IP address but all of the clients run on a single machine. Hopefully this can be useful for others with similar setup.


orgads:
@lemenkov ping

PR #634, opened 2023-05-03
Prevent rtp port exhaustion

atanasdb:


smititelu:
I can argue against this fix with the following: "for" loop for next port will be ran much more loops :D. Please have a look on PR #635

lemenkov:
@atanasdb thanks for the PR and sorry fr the hiatus. Unfortunately we'd better go with PR #635 so I am closing this.

PR #635, opened 2023-05-15
Don't always reset next rtp port to min port

smititelu:
The side effect of this bug is that one can not open more than 100 RTP simultaneous ports to do a load test. This is because of this initial code:
```
    int max_tries = (min_rtp_port < (max_rtp_port - 2)) ? 100 : 1;

```


smititelu:
Checks failed, not related to my commit.

smititelu:
This fix is an alternative for this fix in the PR: #634

prayush-mangal:
yes, your changes fixed the bug. thanks

smititelu:
Rebased this one too. Thanks.

PR #641, opened 2023-06-22
Ims ipsec support

borikinternet:
I was created additional program, which is used to calculate integrity and cyperhing keys before [authentication ...] field will be parsed and auth data will be calculated, so ipsec transport can be configured before sending of new message over encrypted channel. Additionally, I was added support of changing source port on `setdest` action. Some other minor changes (like adding some info into sipp.dtd) implemented too

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,4 +252,6 @@ if(USE_SCTP)
   endif()
 endif()
 
+add_subdirectory(calc_keys)
+
 install(TARGETS sipp DESTINATION bin)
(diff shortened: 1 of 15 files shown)


debayan-ch:
We are looking for IPSec support in sipp and hence wanted to know if this code has been tested against any IMS server with IPSec support.

Thx in advance.
Debayan(Rebaca Technologies).

borikinternet:
Yes, this code was written to test kamailio ims installation and tested with some other lab of one of 4G/LTE mobile phone operator in East Asia, but I don't know really what was installed there.

PR #647, opened 2023-08-23
Use an unordered map for user params

orgads:


lemenkov:
Dupe of #552 ?

orgads:
Yes, sorry

PR #648, opened 2023-08-23
RTPStream: Avoid multi-characters

orgads:


orgads:
@lemenkov

lemenkov:
Looks good.

PR #649, opened 2023-08-23
CMake: Increase minimum version to 3.5

orgads:
CMake is about to drop support for earlier versions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.4)
+cmake_minimum_required(VERSION 3.5)
 
 # set the project name
 project(SIPp)


lemenkov:
I don't see any issues however my opinion you should elaborate why this change is beneficial for SIPp. At least it could cmplicate  SIPp build process on some (unknown to me) old platforms.

lemenkov:
Ok I [looked](https://discourse.llvm.org/t/upgrading-llvms-minimum-required-cmake-version/54879) at a similar proposal and it looks like we can do it quite safely. As a bonus we now can officially use generator expressions within ``INSTALL`` directive. I thing we can try raise it a bit more since as someone in another [issue](https://eigen.tuxfamily.org/bz/show_bug.cgi?id=1767) stated ann CMake versions **higher** than 3.5 have a better CXX11 support.

PR #676, opened 2023-12-17
Support SHA-256 algorithm (RFC 8760)

Maratk1n:


orgads:
Thank you!

1. Please add `-DUSE_SHA256=1` also to build.sh in `--full`.
2. There are many deprecation warnings with openssl 3. Example:

```
src/auth.cpp:348:18: warning: ‘int SHA256_Update(SHA256_CTX*, const void*, size_t)’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  348 |     SHA256_Update( &SHA256Ctx, (unsigned char *) user, strlen(user));
      |     ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Maratk1n:
Hi, @orgads 

1. Done
2. I replaced openssl/sha with [EVP module](https://wiki.openssl.org/index.php/EVP_Message_Digests)

Thanks for review!

orgads:
Thank you!

I didn't test functionality yet, but compilation and tests pass.

build.sh
```diff
@@ -21,7 +21,7 @@ if test "$*" = "--none"; then
 elif test "$*" = "--common"; then
     cmake . -DUSE_GSL=1 -DUSE_PCAP=1 -DUSE_SSL= -DUSE_SCTP=
 elif test "$*" = "--full"; then
-    cmake . -DUSE_GSL=1 -DUSE_PCAP=1 -DUSE_SSL=1 -DUSE_SCTP=1
+    cmake . -DUSE_GSL=1 -DUSE_PCAP=1 -DUSE_SSL=1 -DUSE_SCTP=1 -DUSE_SHA256=1
```
orgads, line 24: Actually why do we need this flag at all? Is there a reason *not* to support it when USE_SSL is set?

orgads:
Verified it works, thank you!

Please remove the flag.

build.sh
Maratk1n: I don't think it's necessary to compile SHA-256 by default because it will always require openssl.

At first I also thought about turning on SHA-256 if `USE_SSL` flag is set, but then I abandoned this idea because these are unrelated things. So I added an extra flag for flexibility.

So, do you think we should enable algorithm by default, or only when `USE_SSL` flag is set?

orgads: I think `USE_SSL` is enough (without it, you can't assume this machine has openssl).

Maratk1n: Done.

orgads:
LGTM

lemenkov:
I've found only a small mostly cosmetic issue. Please address it and rebase on top of current master branch.

Maratk1n:
Hi @lemenkov . I did it, please take a look.

CMakeLists.txt
```diff
@@ -129,10 +129,11 @@ if(USE_SSL)
       message(FATAL_ERROR "Neither OpenSSL nor WolfSSL was found; please install a devel package")
     endif()
   endif()
+  add_definitions("-DUSE_TLS" "-DUSE_SHA256")
```
orgads, line 132: This is bad. Please change line 140 to `if(USE_SSL)`, and move this section there (the existing block should be in `else(USE_SSL)`).

orgads: See https://github.com/orgads/sipp/commit/f30eeb50127ea8b876cc8819dc607959d2b88a40

Maratk1n: @orgads please take a look https://github.com/SIPp/sipp/pull/676/commits/42eac7f0a0c8c65e8d183e771010e70665d68429.

orgads: Still not good. `USE_SSL` can be changed to 0 in this scope.

Maratk1n: Oh, I see. I didn't notice the recent change in master branch
```
message(STATUS "Neither OpenSSL nor WolfSSL was found; disabling sips support")
set(USE_SSL 0)
```
Fixed.

orgads:
Thank you! Maybe squash all the commits at this point? I would leave only the first one as a separate commit.

Maratk1n:
No problem. I've squashed commits and rebased branch.

orgads:
Thanks! Please resolve the open discussion.

Maratk1n:
Did you mean comment from [github-advanced-security](https://github.com/apps/github-advanced-security) bot?
I fixed failure but don't have permission to resolve the thread.

orgads:
I meant our discussion about CMakeLists. You already resolved it :)

lemenkov:
> Did you mean comment from [github-advanced-security](https://github.com/apps/github-advanced-security) bot? I fixed failure but don't have permission to resolve the thread.

@Maratk1n your PR raises a minimum OpenSSL version up to 1.1.0 but still there is a mention of 0.9.8 version in the doc-files. See my notes above.

Maratk1n:
Thank you @lemenkov . Fixed.

lemenkov:
@orgads looks good to me. What's your opinion?

orgads:
lgtm

PR #677, opened 2023-12-28
More gh actions

lemenkov:


lemenkov:
@rkday what do you think?

lemenkov:
Ok let's merge it. It passes the tests and we didn't see any news from Coverity since 2016.

PR #682, opened 2024-01-12
Build: Auto-detect available libraries by default

orgads:
If neither OFF or ON is defined, enable if auto-detected.


lemenkov:
LGTM

Issue #632, opened 2023-04-18
How to send comfort noise or make sipp read and send the comfort noise from pcap file ?


Costallat:
Hi all,

I'm trying to reproduce a problem here using sipp to send the same packets(pcap) where we saw the problem. I already made it read the pcap file and send the packets, I can hear the RTP stream(audio), but we think that what is causing the problem is two comfort noise packets that are being sent by the server, but I saw that sipp is not sending those packet, it only sends the 'audio' RTP packets. 

I'm trying to check the code here to see if there is any kind of filter to get only RTP packets with specific payload type where I can add the comfort noise payload so sipp can send them.

Does anybody have any clue or tips where I can check for that? Or if there is another way to generate those comfort noise packets?

Thanks

lemenkov:
PR #687 was merged so it should work now.

PR #687, opened 2024-01-16
Add media support for comfort noise (audio/CN)

rsvargas:
Make it possible to playback a PCAP file or echo received streams containing CN packets.


rsvargas:
This will also fix #632.

lemenkov:
Looks good to me. Passes all the tests. One question - have you tested it?

rsvargas:
I did tested it, but it was some time ago, I've had this commit in my local copy for a while, so when I saw it recently I rebased my branch and made the PR.

lemenkov:
OK, let's merge it.

Issue #576, opened 2022-04-08
segfault in send_packets


davehorton:
Built the latest from source, its crashing on me.

```
Core was generated by `sipp -sf uac_pcap_60s_call.xml -r 20 -m 500 -l 300 -s 555 3.120.157.16'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000563daf999838 in send_packets (play_args=0x563db050dff0) at /usr/local/src/sipp/src/send_packets.c:261
261	        memcpy(udp, pkt_index->data, pkt_index->pktlen);
[Current thread is 1 (Thread 0x7f6dffd1c700 (LWP 10302))]
(gdb) bt
#0  0x0000563daf999838 in send_packets (play_args=0x563db050dff0) at /usr/local/src/sipp/src/send_packets.c:261
#1  0x0000563daf9652b4 in send_wrapper (arg=0x563db050dff0) at /usr/local/src/sipp/src/call.cpp:6818
#2  0x00007f6e21573fa3 in start_thread (arg=<optimized out>) at pthread_create.c:486
#3  0x00007f6e21133eff in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb) l
256	     * allows the thread to be cancelled cleanly.
257	     */
258	    pthread_cleanup_push(send_packets_cleanup, ((void *) &sock));
259
260	    while (pkt_index < pkt_max) {
261	        memcpy(udp, pkt_index->data, pkt_index->pktlen);
262	        port_diff = ntohs(udp->uh_dport) - pkts->base;
263	        /* modify UDP ports */
264	        udp->uh_sport = htons(port_diff + ntohs(*from_port));
265	        udp->uh_dport = htons(port_diff + ntohs(*to_port));
```

latest master branch
```
f44d0cf (HEAD -> master, origin/master, origin/HEAD) docs: Change scenario in CHANGES.md to one that works
```

is this a known issue?

davehorton:
perfectly recreatable under load by the way

MichaelSekora:
Hi Dave,
I try to reproduce the scenario, could you please provide "uac_pcap_60s_call.xml", I cannot find it in the example scenarios, etc.
Do you have a sip-trace? how does the other side respond (connected with 200 OK)?

Issue #673, opened 2023-12-11
segmentation fault on playing pcap file ( sipp 3.7.2 )


gitkoyot:
Hello,
I have a scenario where both UAC and UAS are playing 10 sec audio. I noticed that sometimes UAS part crashes. When I executed it with valgrind this is what I found:

```
==6613== Thread 2:
==6613== Invalid read of size 2
==6613==    at 0x185541: send_packets (send_packets.c:241)
==6613==    by 0x13936F: send_wrapper(void*) (call.cpp:6824)
==6613==    by 0x53183EB: start_thread (pthread_create.c:444)
==6613==    by 0x539896F: clone (clone.S:100)
==6613==  Address 0x647d988 is 8 bytes inside a block of size 40 free'd
==6613==    at 0x48431EF: free (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==6613==    by 0x1170B6: CAction::setPcapArgs(char const*) (actions.cpp:536)
==6613==    by 0x136152: call::executeAction(char const*, message*) (call.cpp:6083)
==6613==    by 0x12456E: call::executeMessage(message*) (call.cpp:1900)
==6613==    by 0x1257E4: call::run() (call.cpp:2214)
==6613==    by 0x1A69F1: traffic_thread(int&, int&) (sipp.cpp:589)
==6613==    by 0x1AB53C: main (sipp.cpp:2147)
==6613==  Block was alloc'd at
==6613==    at 0x4840808: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==6613==    by 0x1170DA: CAction::setPcapArgs(char const*) (actions.cpp:541)
==6613==    by 0x136152: call::executeAction(char const*, message*) (call.cpp:6083)
==6613==    by 0x12456E: call::executeMessage(message*) (call.cpp:1900)
==6613==    by 0x1257E4: call::run() (call.cpp:2214)
==6613==    by 0x1A69F1: traffic_thread(int&, int&) (sipp.cpp:589)
==6613==    by 0x1AB53C: main (sipp.cpp:2147)
```

also stack trace points to similar
```
                                                                                                                                        Thread 11 "sipp" received signal SIGSEGV, Segmentation fault.
                                                        [Switching to Thread 0x7ffff24bd6c0 (LWP 7449)]
 ...

(gdb) bt full
[...]

Issue #610, opened 2023-02-24
SIPp giving segmentation fault.


sahil-ecosmob:
Hello. I am trying to do the load balancer test using SIPp. I am using the following commands for UAC and UAS : 

run UAS : 
sipp -sf Active-Media/uas-active.xml -i 192.168.1.10 -p 5064 -rtp_echo -m 500

run UAC : 
sipp -sf Active-Media/uac-active.xml -i 192.168.1.10 192.168.1.10:5064 -rtp_echo -m 500 -r 100

It works fine until -r flag has value < 50. above this, it shows segmentation fault. Also tI don't face this problem with basic scenarios... only when I use RTP. could it be because of RTP port allocation or something ??

Any idea why this could be happening ?? 
Thanks in advance for help. 

Regards Sahil

wdoekes:
Those scenarios are not included in this repo. Maybe you can attach them or add them as `<details><pre>` so people can reproduce.

sahil-ecosmob:
sure. I am using scenarios from this git repo. https://github.com/pbertera/SIPp-by-example

wdoekes:
That's not attaching. If someone decided to look at this in a month, that repo might be looking differently. Please make it easy others.

sahil-ecosmob:
Hello. These are the scenario files. I have added them as .txt files as git wasn't allowing .xml format. So please rename to .xml whenever used. 

[uac-active.txt](https://github.com/SIPp/sipp/files/10822051/uac-active.txt)
[uas-active.txt](https://github.com/SIPp/sipp/files/10822052/uas-active.txt)

rafiw:
I also got a segmentation fault when using r=100, I used valgrind and GDB to track the problem here is the trace
the segmentation happens in this line in send_packets
```
   while (pkt_index < pkt_max) {
        memcpy(udp, pkt_index->data, pkt_index->pktlen); // crash here
```
using gdb i can see pkt_index is null. it looks like thread sync issue.
```
==29937== Thread 2:
==29937== Invalid read of size 8
==29937==    at 0x206447: send_packets (send_packets.c:243)
==29937==    by 0x1A02C7: send_wrapper(void*) (call.cpp:6820)
==29937==    by 0x54966DA: start_thread (pthread_create.c:463)
==29937==    by 0x6E0661E: clone (clone.S:95)
==29937==  Address 0x77df840 is 32 bytes inside a block of size 40 free'd
[...]

PR #521, opened 2021-04-12
RTPStream/PCAP: Support variables for filename

orgads:
Allow using `[fieldX]` and all types of variables on playback.

PR #698, opened 2024-03-20
Remove support for variable PCAP filenames

peter-oneill:
Fixes issues #576, #610, #673

The original variable PCAP filenames PR #521 would free and rewrite a single per-scenario-action block of memory for every call, but all calls would continue referencing the memory.  This led to various use-after-free issues.  The PCAP play design also isn't set up for loading a file per call, so the solution didn't scale well.

This PR removes the variable PCAP filenames feature, while retaining the variable RTPStream filenames feature, which doesn't have the same issues.

Specific testing done:
- that PCAP play is fixed
- that RTPStream continues to work, with files specified as any of:
  - `<exec rtp_stream="my_audio.wav"/>`
  - `<exec rtp_stream="[field0]"/>`
  - `<exec rtp_stream="[command_line_key]"/>`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,13 @@
+BREAKING(!) changes in 3.7.3
+============================
+
+- Remove support for variables in PCAP filenames, originally introduced in 3.7.0~rc1.  See #673
+    
 Bugs fixed in 3.7.3
 ===================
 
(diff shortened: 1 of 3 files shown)


orgads:
Thank you very much @peter-oneill!

Please add a release note.

PR #703, opened 2024-03-25
Get rid of "using namespace std" everywhere

orgads:
Use consistent scoping.


orgads:
@wdoekes The failures here are unrelated to my changes. Can you please approve it?

wdoekes:
No sneaky ifunc additions here, right? ;P

orgads:
🤣🤣

wdoekes:
There are some other changes than just adding std:: (adding auto, replacing heap with stack alloc), but I guess it's fine. We did auto already, I see. And the stack is only for the better here.

wdoekes (approved):

orgads:
I preferred changing to auto rather than `std::map<whatever>::iterator`...

PR #704, opened 2024-03-25
Replace NULL with nullptr

orgads:


wdoekes:
I don't c++ on a regular basis. This `new(std::nothrow)` business is new (haha) to me.

If all `new`'s are supposed to `throw bad_alloc` then most of the c++ style code allocation is broken.

I don't see any `-fno-exceptions` anywhere, and I don't see any explicit `operator new` override using malloc() either. So I assume normal `new` failures will return in an unhandled exception and an abort().

orgads:
@wdoekes I marked them as won't fix to silence the checker, and pushed a fixing follow-up commit to https://github.com/orgads/sipp/tree/less-null-validations.

I'll open a PR after this one is merged, since it is a dependency.

wdoekes (approved):
Not sure if this replacement was the first on my list. But more c++ is better.

PR #708, opened 2024-04-05
Avoid GNU-specific xargs flags in regress/runtests

ryandesign:
In fact, avoid `xargs` entirely by using features built into `find` instead.


orgads:
Thanks!

Issue #711, opened 2024-04-05
/usr/bin/pager: No such file or directory


ryandesign:
After compiling the latest code in master on macOS, running `./sipp` produced this error:

```
/usr/bin/pager: No such file or directory
```

followed by several pages of usage information.

There is no `pager` on macOS. There are, of course, `less` and `more`.

wdoekes:
Bah. I was hoping to avoid having to add multiple options there, assuming that pager was ubiquitous.

We'd have to add multiple options here then:
```c
    if (!argv[0]) {
        argv[0] = pager; /* missing PAGER */
```
Moving the `access()` call up to check for each possibility and then simply bailing without (p)error if none are found.

PR #714, opened 2024-04-05
Try less and more when pager is not found (on macOS?)

wdoekes:
Closes: #711
Reported-by: Ryan Carsten Schmidt

### DIFF
--- a/src/sipp.cpp
+++ b/src/sipp.cpp
@@ -756,7 +756,15 @@ static char* wrap(const char* in, int offset, int size)
 /* If stdout is a TTY, wrap stdout in a call to PAGER (generally less(1)).
  * Returns a pid_t you'll have to pass to end_pager(). */
 static pid_t begin_pager() {
-    char pager[15] = "/usr/bin/pager";
+    char const *pager_options[4] = {
+        "/usr/bin/pager",
+        "/usr/bin/less",
+        "/usr/bin/more",
+        nullptr
+    };
+    char const *const *pptr;
+    char const *ptr;
+    char argv0_buf[PATH_MAX + 1];
     char *argv[2] = {nullptr, nullptr};
 
     int stdout_fd = fileno(stdout);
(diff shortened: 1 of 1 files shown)


src/sipp.cpp
```diff
@@ -767,17 +775,33 @@ static pid_t begin_pager() {
...
-    /* Get pager first, so we can bail if it's not there */
-    argv[0] = getenv("PAGER");
-    if (!argv[0]) {
-        argv[0] = pager; /* missing PAGER */
-    } else if (!*argv[0]) {
-        return 0; /* blank PAGER */
+    /* Get pager first, so we can bail if it's not there. Prefer env */
+    if ((ptr = getenv("PAGER"))) {
+        if (!ptr[0]) {
+            return 0; /* blank PAGER */
```
orgads, line 781: !ptr

src/sipp.cpp
```diff
@@ -756,7 +756,15 @@ static char* wrap(const char* in, int offset, int size)
 /* If stdout is a TTY, wrap stdout in a call to PAGER (generally less(1)).
  * Returns a pid_t you'll have to pass to end_pager(). */
 static pid_t begin_pager() {
-    char pager[15] = "/usr/bin/pager";
+    char const *pager_options[4] = {
```
orgads, line 759: `std::vector<std::string>`? Iteration is easier, and you won't need a copy, you can pass `&opt[0]`.

orgads (changes requested):

src/sipp.cpp
wdoekes: No. `!ptr[0]` or `ptr[0] == '\0'`

The `if` at 779 already checks `!ptr`.

src/sipp.cpp
wdoekes: :shrug: 

I guess we should do more c++. Be my guest :grimacing: 

I assume we won't need a copy because it _is_ a copy already.

orgads (approved):
You can merge this as it is, I'll C++ify it later :)

wdoekes:
closed by 3242112e3fb0d003b15d8ddaade4d468f01068f1

PR #724, opened 2024-04-22
Fix issue where next_url could return garbage

mstovenour:
next_req_url, which is a zero terminated string, is not initialized when initially allocated.  Later code correctly zero terminates this string 99% of the time but fails to do so if there are no Record-Route or Contact headers.  Subsequent use of [next_url] will return garbage, best case.


wdoekes (approved):
Brr.. how that next_req_url is allocated does not win any prizes.

But you're not making it worse :+1: 

Thanks!

PR #725, opened 2024-04-22
Further explain xml syntax change for 3.6.0

mstovenour:


docs/scenarios/actions.rst
```diff
@@ -125,6 +125,9 @@ The following example is used to:
       </action>
     </recv>
 
+.. note::
+  Release 3.6.0 added XML syntax checks and the characters & and < are
+  not valid in XML files.  Replace those with &amp; &lt; in strings.
```
wdoekes, line 130: I would say:
```
  Release 3.6.0 added rudimentary XML syntax checks. Now the &
  and < characters must be escaped even inside attribute values.
```

docs/scenarios/variables.rst
```diff
@@ -59,10 +59,12 @@ clause. For example:
...
-        <ereg regexp="<sip:([^;@]*)" search_in="hdr" header="To:" assign_to="dummy,uri" />
+        <ereg regexp="&lt;sip:([^;@]*)" search_in="hdr" header="To:" assign_to="dummy,uri" />
       </action>
     </recv>
     <Reference variables="dummy" />
 
-
+.. note::
+  Release 3.6.0 added XML syntax checks and the characters & and < are
+  not valid in XML files.  Replace those with &amp; &lt; in strings.
```
wdoekes, line 69: same

src/scenario.cpp
```diff
@@ -1026,7 +1026,7 @@ scenario::scenario(char * filename, int deflt)
     /* Close scenario element */
     xp_close_element();
     if (xp_is_invalid()) {
-        ERROR("Invalid XML in scenario");
+        ERROR("Invalid XML in scenario.  Check for & and < in strings; replace with &amp; &lt;");
```
wdoekes, line 1029: Maybe:
```
Invalid XML in scenario. See: https://github.com/SIPp/sipp/issues/414
```
Because there could certainly be other reasons than unescaped lt and amp.

wdoekes (changes requested):
Thanks for the PR :)

docs/scenarios/actions.rst
mstovenour: Can we explain how to escape them?  Again I expect a lot of people to trip over this and many of them are not programmers and may have only ever used XML with SIPp.  e.g. my entire SVT test organization.

  Release 3.6.0 added rudimentary XML syntax checks. Now the &
  and < characters must be escaped (&amp; &lt;) even inside attribute values.

wdoekes: We can. How about:
```
Release 3.6.0 added rudimentary XML syntax checks. Now the & and <
characters must be escaped as &amp; and &lt; even inside attribute values.
```

src/scenario.cpp
wdoekes: Here's a commit/branch that adds line numbers in _some_ cases:

https://github.com/SIPp/sipp/commit/a3b6289b2fabf1351d43e050709290b948939144

It's not pretty, but it's fairly non-intrusive.

Please check it this meets your needs.

wdoekes:
Looks good. Can you squash these changes?

mstovenour:
I fixed that.  Sorry about all the noise.  I was editing the docs directly in github and they make a commit for each file edit.

wdoekes:
Thanks!

Issue #730, opened 2024-05-16
The `-mp` option is still available (good) but not documented (bad)


gtjoseph:
3.6.0 removed the `-min_rtp_port` and `-max_rtp_port` options and required the [Asterisk](https://github.com/asterisk) team to modify all of the testsuite tests that used those options to use `-mp` instead.  3.7.0, restored the `-min_rtp_port` and `-max_rtp_port` options (thank you) but appeared to remove the `-mp` option requiring us to change back again.  I was about to submit a PR to restore `-mp` but it looks like the functionality is still there (setting min_rtp_port) just the help text was removed.

So first... We've got almost 500 tests that use sipp so messing with the options is an issue for us.  If you do need to remove one, at least give us some notice.  It looks like the [sourceforge mailing list](https://sourceforge.net/p/sipp/mailman/sipp-users/) is dead.  Is there an active announcement type replacement?

Second... There's a PR incoming to restore the help text for `-mp`.

Thanks

PR #731, opened 2024-05-16
sipp.cpp:  Restore help for the -mp option

gtjoseph:
Set help text for the `-mp` option to...
"Sets -min_rtp_port for backwards compatibility."

Fixes: #730


orgads:
Thank you!

PR #734, opened 2024-06-25
Fix rtpstream local port allocation

viktike:
Don't hardcode retry counter max value.

After 100 calls with RTP the following error was shown before:

```
2024-06-25      07:35:06.156700 1719300906.156700: Could not bind port for RTP streaming after 100 tries
2024-06-25      07:35:06.156776 1719300906.156776: cannot assign a free audio port to this call - using 0 for [rtpstream_audio_port]
```


orgads:
Thanks!

PR #738, opened 2024-08-01
Initialize srand with hostname and PID

FalacerSelene:
We're running SIPp in a container scaleset, and ran into an issue where two instances started in the same second and continually picked the same random numbers as each other, which meant that `InputFileRandomOrder` actually had the instances continually 'randomly' picking the same entries.

To avoid this, this change adds a bit more noise to the start-of-day call to `srand()`, mixing in the PID and the hostname, which would avoid two instances using the same seed.

I'd welcome an alternative and better way to do this!

### DIFF
--- a/src/sipp.cpp
+++ b/src/sipp.cpp
@@ -1334,6 +1334,48 @@ static void setup_media_sockets()
     }
 }
 
+/* Seed the random number generator from a combination of
+ * - Nanoseconds since epoch
+ * - Process ID
+ * - Hostname
+ */
+void randomseed(void)
+{
+    struct timespec ts;
+    unsigned int    seed          = time(nullptr);
+    char            hostname[256] = { 0 };
+    pid_t           p             = getpid();
+    int             index;
+
+    // If the system clock can provide us with a nanosecond count, then
+    // include that in the seed.
+    if (clock_gettime(CLOCK_MONOTONIC, &ts) == 0) {
+        seed ^= (unsigned int)ts.tv_nsec;
+    }
+
+    // The system clock might not have nanosecond resolution, and so we still
+    // want to combine the seed with other values to avoid two
+    // near-simultaneous launches getting the same random seed.
+
+    if (p != 0) {
+        // We don't XOR with the PID because if we launch twice successively
+        // it's possible the PID increment could precisely counter the time
+        // increment.
+        seed *= p;
+    }
+
+    // Might truncate, or might leave 0s after the end.
+    gethostname(hostname, 256);
+
+    for (index = 0; index < 256; index++) {
+        // Bitflip with successive bytes of the hostname. Once we get to the
+        // 0s at the end it's a noop.
+        seed ^= (hostname[index] << (8 * (index % sizeof(seed))));
+    }
+
+    srand(seed);
+}
+
 /* Main */
 int main(int argc, char *argv[])
 {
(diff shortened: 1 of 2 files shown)


src/sipp.cpp
```diff
@@ -1334,6 +1334,38 @@ static void setup_media_sockets()
...
+ */
+void randomseed(void)
+{
+    time_t       t             = time(nullptr);
+    pid_t        p             = getpid();
+    unsigned int seed          = t;
+    char         hostname[256] = { 0 };
+    int          index;
+
+    if (p != 0) {
```
FalacerSelene, line 1350: This ought to never be 0 but better safe than sorry.

src/sipp.cpp
```diff
@@ -1334,6 +1334,38 @@ static void setup_media_sockets()
...
+        // We don't XOR with the PID because if we launch twice successively
+        // it's possible the PID increment could precisely counter the time
+        // increment.
+        seed *= p;
+    }
+
+    // Might truncate, or might leave 0s after the end.
+    gethostname(hostname, 256);
+
+    for (index = 0; index < 256; index++) {
```
FalacerSelene, line 1360: I also could have done `gethostname(.., 255)` and iterated here while `hostname[index] != '\0'`, but I didn't see much in it either way.

orgads:
Claude suggested this:
```c++
unsigned int generate_seed() {
    unsigned int seed = (unsigned int)time(NULL);
    seed ^= (unsigned int)clock();
    seed ^= (unsigned int)getpid();
    
    struct timespec ts;
    if (clock_gettime(CLOCK_MONOTONIC, &ts) == 0) {
        seed ^= (unsigned int)ts.tv_nsec;
    }
    
    return seed;
}
```

WDYS?

orgads:
> > Claude suggested this:
> > ```c++
> > unsigned int generate_seed() {
> >     unsigned int seed = (unsigned int)time(NULL);
> >     seed ^= (unsigned int)clock();
> >     seed ^= (unsigned int)getpid();
> >     
> >     struct timespec ts;
> >     if (clock_gettime(CLOCK_MONOTONIC, &ts) == 0) {
> >         seed ^= (unsigned int)ts.tv_nsec;
> >     }
> >     
> >     return seed;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > WDYS?
> 
> I don't think we gain from mixing 3 different time-based values (`time()`, `clock()` and `clock_gettime()`) - that still gives only time-based uniqueness. So I'd probably just take `clock_gettime()`'s nanoseconds and forget about `time()` and `clock()` in that case, because it's incredibly unlikely that two processes call that in the same ns.
> 
> However, `clock_getres()` might indicate the resolution isn't really to the nanosecond. Since we don't know the resolution of the clock across systems that sipp might run on, I thought it was better to grab other sources of uniqueness (PID and hostname).
> 
> But a suggestion like this would almost definitely solve my problem too :)

Sounds good. So can you implement it?

FalacerSelene:
> Sounds good. So can you implement it?

Yep, updated to XOR with the nanosecond value too :)

orgads:
Any idea why the test fails?

orgads:
@jeannotlanglois ?

orgads:
Rebased and it passed. Merging.

Thanks for your persistence @FalacerSelene!

PR #754, opened 2024-09-11
Include the SNI for SSL client

### LINKED ISSUE #753: Missing TLS extension support for Server name indication (SNI) in client mode
When running SIPp as a client, we can provide the server endpoint. When using the TLS transport and a hostname server endpoint, the 
application already resolve the ip of the host. However, there are no SNI added in the ClientHello during the TLS handshake.

This is something supported by OpenSSL than can be programmatically injected into the ssl context. I guess that WolfSSL supports it as well, but have not checked.

Do you see this as a good addition to the application ?
It could be either compilation flags (like Kamailio) or execution options.

Relevant links : 
- https://www.cloudflare.com/learning/ssl/what-is-sni/
- https://datatracker.ietf.org/doc/html/rfc6066#section-3
- https://manpages.debian.org/experimental/libssl-doc/SSL_set_tlsext_host_name.3ssl.en.html
- Some known opensource server also support it : https://github.com/search?q=repo%3Akamailio%2Fkamailio%20tlsext&type=code

grondinjc:
Add SNI for SSL client only.
As SSL related function applies any input into the SNI extension, a minimal validation was added to ignore IP endpoint.

From wireshark, the TLS handshake has visible differences
image
- The 1st clientHello is the result of the command `sipp -t l1 -sn uac <server_as_hostname>:5061`
- The 2nd clientHello is the result of the command `sipp -t l1 -sn uac <server_as_direct_ip>:5061`

fixes #753

### DIFF
--- a/src/sslsocket.cpp
+++ b/src/sslsocket.cpp
@@ -126,7 +126,16 @@ const char *SSL_error_string(int ssl_error, int orig_ret)
 
 SSL* SSL_new_client()
 {
-    return SSL_new(sip_trp_ssl_ctx_client);
+    SSL* ssl = SSL_new(sip_trp_ssl_ctx_client);
+
+    // Inject the ServerNameIndication (SNI).
+    // It requires to be a hostname. However, SSL takes whatever we set.
+    // Ref https://datatracker.ietf.org/doc/html/rfc6066#section-3
+    if (strcmp(remote_ip, remote_host)) {
+        SSL_set_tlsext_host_name(ssl, remote_host);
+    }
+
+    return ssl;
 }
 
 SSL* SSL_new_server()


src/sslsocket.cpp
```diff
@@ -126,7 +126,16 @@ const char *SSL_error_string(int ssl_error, int orig_ret)
 
 SSL* SSL_new_client()
 {
-    return SSL_new(sip_trp_ssl_ctx_client);
+    SSL* ssl = SSL_new(sip_trp_ssl_ctx_client);
+
+    // Inject the ServerNameIndication (SNI).
+    // It requires to be a hostname. However, SSL takes whatever we set. 
+    // Ref https://datatracker.ietf.org/doc/html/rfc6066#section-3
+    if (strcmp(remote_ip, remote_host) != 0){
```
orgads, line 134: I prefer `!strcmp(...)`, and please add a space before the brace.

orgads:
Thank you!

src/sslsocket.cpp
grondinjc: on it

grondinjc: in my case, `!strcmp` would not be right. I need to set the hostname if it differs from the remote_ip

orgads: Right, my mistake.

src/sslsocket.cpp
```diff
@@ -126,7 +126,16 @@ const char *SSL_error_string(int ssl_error, int orig_ret)
 
 SSL* SSL_new_client()
 {
-    return SSL_new(sip_trp_ssl_ctx_client);
+    SSL* ssl = SSL_new(sip_trp_ssl_ctx_client);
+
+    // Inject the ServerNameIndication (SNI).
+    // It requires to be a hostname. However, SSL takes whatever we set.
```
orgads, line 132: You have a trailing WS here.

orgads:
Please squash the commits.

src/sslsocket.cpp
grondinjc: just noticed from the linting step. done

PR #757, opened 2024-09-15
SRTP: Replace AES ECB with AES CBC

orgads:
ECB is not built on alpine wolfssl.


orgads:
@jeannotlanglois Can you suggest what to do here? This breaks the srtp test.

orgads:
It also fails in #738.

jeannotlanglois:
As per SIP RFCs 3711 and 4568, ECB is needed in order to use the counter mode (CTR) which is *required* by the SRTP protocol.
As far as my understanding goes, this cannot be changed as the protocol is based on that AES behavior.  This is why the SRTP tests fail.

orgads:
Thank you @jeannotlanglois. I'll see what I can do.

orgads:
Superseded by #759.

PR #763, opened 2024-10-07
Add support for raw IP pcap files

ticpu:
Files created from Wireshark by using "File -> Strip Headers" and that contain only the minimum amount of information to send RTP packets. No headers to strip in this case so set offset to 0.

Documentation for DLT_RAW / raw IP / link layer 12 is available at: https://github.com/the-tcpdump-group/libpcap/blob/master/pcap/dlt.h


orgads:
LGTM.

ticpu:
I did some testing on my side and was able to get some calls using v4 and v6 using a raw capture.

PR #766, opened 2024-10-10
Make <nop> override the call::last_action_result only if it is a failure

tghosgor:
Otherwise it clears the failure flag from the call even if a previous action failed.


orgads:
Merged, thanks!

PR #769, opened 2024-11-04
Added regexp support for response codes

petr-cisar:


petr-cisar:
Now the regexp_match attribute should also work for response codes. Example:
`<recv response="^18" regexp_match="true" optional="true">`

orgads:
Nice! Can you please add a test?

src/call.cpp
```diff
@@ -4451,24 +4451,45 @@ bool call::matches_scenario(unsigned int index, int reply_code, char * request,
...
-        /* This is a potential candidate, we need to match transactions. */
-        if (curmsg->response_txn) {
-            if (transactions[curmsg->response_txn - 1].txnID && !strcmp(transactions[curmsg->response_txn - 1].txnID, txn)) {
-                return true;
-            } else {
-                return false;
+    } else if (curmsg->recv_response) {
+        if (curmsg->regexp_match) { // Match response code using regex
+            char reply_code_str[8];
+           snprintf(reply_code_str,8,"%u",reply_code); // Convert the response code to string
```
orgads, line 4457: Spaces after comma.

src/call.cpp
```diff
@@ -4451,24 +4451,45 @@ bool call::matches_scenario(unsigned int index, int reply_code, char * request,
...
-            /* Always true for the first message. */
-            return true;
-        } else if (curmsg->recv_response_for_cseq_method_list &&
-                   strstr(curmsg->recv_response_for_cseq_method_list, responsecseqmethod)) {
-            /* If we do not have a transaction defined, we just check the CSEQ method. */
-            return true;
-        } else {
-            return false;
-        }
+            if (regexec(curmsg->regexp_compile, reply_code_str, (size_t)0, nullptr, REGEXEC_PARAMS)) {
```
orgads, line 4468: Is the cast really needed?

src/call.cpp
```diff
@@ -4451,24 +4451,45 @@ bool call::matches_scenario(unsigned int index, int reply_code, char * request,
...
+               return false;
+           }
+        } else { // Exact numerical match
+            if (atoi(curmsg->recv_response)!=reply_code) {
```
orgads, line 4472: Spaces around `!=`

orgads (commented):
Please rebase.

regress/github-#0769/run
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+# This regression test is a part of SIPp.
+# Author: Walter Doekes, OSSO B.V., 2015
```
orgads, line 3: Change it to your name :)

orgads (changes requested):

orgads:
The second commit message mentions style fixes and DTD update, but looks like you forgot to stage them?

src/call.cpp
petr-cisar: I have just copy/pasted this section of code from the request handling few lines above. So I would suggest either to keep both, or remove it at both places. I guess that it has no impact, so it is just matter of style.

orgads:
Great, just fix the test author name and squash the commits and we're good to go.

petr-cisar:
Commits squashed. Hopefully all is OK, this is my first attempt to contribute to a project on GH, so it takes some googling and learning.

orgads (approved):

orgads:
You lost the merged changes on squash. I fixed it for you.

orgads:
Thank you!

PR #772, opened 2025-01-16
Set null terminator after realloc

rajeshsingh381:
When we `realloc` for a smaller message than previous, it sometimes detects old SIPp records from old messages making the record malformed. For example, `call-id` parsing gets malformed in new smaller message because it matches 2 lines, due to missing null terminator.

TO-DO: fix all `realloc` instances


orgads:
Thanks for pushing this. Do you plan to complete it?

PR #781, opened 2025-03-11
TLS key logging

g-eye42:
TLS key logging records the TLS Session Keys to a key log file when the `SSLKEYLOGFILE` environment variable is set. It allows to decrypt SIPS traffic generated or received by SIPp using Wireshark. For more details see: https://wiki.wireshark.org/TLS
The TLS key log file format is described here: https://datatracker.ietf.org/doc/draft-ietf-tls-keylogfile/


src/sslsocket.cpp
```diff
@@ -397,6 +453,13 @@ enum tls_init_status TLS_init_context(void)
         return TLS_INIT_ERROR;
     }
 
+#ifdef HAVE_KEYLOG_CB
+    if (keylog_file()) {
+        SSL_CTX_set_keylog_callback(sip_trp_ssl_ctx_client, sip_tls_keylog_callback);
```
orgads, line 458: Is it needed also for `sip_trp_ssl_ctx`?

g-eye42: In order to be able to use the key logging also in the server role, the callback has to be registered for `sip_trp_ssl_ctx` as well...

orgads:
Looks good, thank you.

Can you please squash all the commits except the cmake one?

g-eye42:
I hope I got you right, please check.

orgads:
I now noticed that the cmake change relates to the first commit, so I just squashed everything. I also wrapped the commit message.

LGTM. Thanks!

PR #786, opened 2025-05-22
fix: force exit when q is pressed twice

luiz-telles:


orgads:
Thanks for the PR. Can you please check if #788 works for you? Do you find any advantages for your fix over that one?

orgads:
Thank you! Merged with a few tweaks.

PR #800, opened 2025-07-29
Prevent warning with epoll when variable is set but not used.

jkroonza:
.../src/socket.cpp:2890:13: error: variable events set but not used [-Werror=unused-but-set-variable=]
 2890 |         int events = 0;

Reference: https://bugs.gentoo.org/960831


jkroonza:
> Thanks!

Pleasure, but primarily it's Gentoo work that runs a tinderbox to find these issues.

PR #817, opened 2025-11-18
Integrated support of WSS transport using libwebsocket

neutrino38:
Adding support of SIP over WSS based on libwebsocket

### DIFF
--- a/src/socket.cpp
+++ b/src/socket.cpp
@@ -944,16 +944,13 @@ void SIPpSocket::invalidate()
         return;
     }
 
+    ss_invalid = true;
+
 #if defined(USE_OPENSSL) || defined(USE_WOLFSSL)
     if (SSL *ssl = ss_ssl) {
         SSL_set_shutdown(ssl, SSL_SENT_SHUTDOWN|SSL_RECEIVED_SHUTDOWN);
         SSL_free(ssl);
     }
-#endif
-
-#ifdef USE_WSS
-    // Close websocket connection
-    close_wss();
 #endif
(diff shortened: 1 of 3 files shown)


neutrino38:
> Hello there!
> 
> Great work, it would be very nice if we could somehow change the ws `ccinfo.path` (For example I have `/ws` on my side) aswell as be able to use flags like `LCCSCF_ALLOW_SELFSIGNED`, `LCCSCF_ALLOW_INSECURE`, `LCCSCF_SKIP_SERVER_CERT_HOSTNAME_CHECK`

regarding the ccinfo.path, how would you specify that in the sipp scenario?
regarding the flags, do you want me to add options to the command line? I'll check how it's done with TLS.

orgads:
@lemenkov Can you please review this?

src/socket.cpp
```diff
@@ -1239,6 +1269,258 @@ void process_message(SIPpSocket *socket, char *msg, ssize_t msg_size, struct soc
...
+
+    struct lws_context_creation_info info = {0};
+    info.port = CONTEXT_PORT_NO_LISTEN;
+    info.protocols = protocols;
+    info.gid = -1;
+    info.uid = -1;
+    info.options = LWS_SERVER_OPTION_DO_SSL_GLOBAL_INIT;
+    info.timeout_secs = 40;
+    info.ka_time = 15;      
+    info.ka_probes = 3;
```
lemenkov: Remove trailing whitespace to make lint happy

PR #822, opened 2025-12-06
SRTP: Fix mis-parsing of long input line

orgads:
A literal space in sscanf means "0 or more spaces". So a string longer than 40 chars was split between e.g. `primary_audio_cryptokeyparams` and `crypto_audio_sessionparams`.

### DIFF
--- a/src/call.cpp
+++ b/src/call.cpp
@@ -452,6 +452,8 @@ int call::extract_srtp_remote_info(const char * msg, SrtpAudioInfoParams &pA, Sr
     std::size_t amsection_limit = 0;
     std::size_t vmsection_limit = 0;
 
+    *crypto_audio_sessionparams = 0;
+    *crypto_video_sessionparams = 0;
     /* Look for start of message body */
     ro_search= strstr(msg, "\n\n"); // UNIX line endings (LFLF) between header/body sections
     alt_search= strstr(msg, "\r\n\r\n"); // DOS line endings (CRLFCRLF) between header/body sections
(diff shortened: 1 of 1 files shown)


jeannotlanglois:
I will take a look on Subnday 
I've just became a first-time new dad recently, time is a bit scarce but planning some on Sunday for a good check on this.  Planning to restore the original code with memsets which had been extensively tested but somehow taken out by soneone.  JLSRTP shouldn't have to be modified as by design caller should provide NULL-terminated values and this has been confirmed to work for many years without a hitch before it was merged in.

orgads:
@jeannotlanglois I replaced memset with assignment of null in position 0. There's no reason to fill the entire buffer just for null-termination. It should just be handled correctly.

Regarding JLSRTP, the only change I did there was to reserve buffer size before appending, to avoid repeating reallocations. It's not critical, but can improve performance a bit.

src/call.cpp
```diff
@@ -555,7 +557,7 @@ int call::extract_srtp_remote_info(const char * msg, SrtpAudioInfoParams &pA, Sr
             mline_eol = msgstr.find("\n", mline_sol, 1);
             if (mline_eol != std::string::npos) {
                 mline_contents = msgstr.substr(mline_sol, mline_eol);
-                sscanf(mline_contents.c_str(), "\na=crypto:%d %24s inline:%40s %63s",
+                sscanf(mline_contents.c_str(), "\na=crypto:%d %24[^ ] inline:%40[^ ]%*1[ ]%63s",
```
jeannotlanglois: Alright - I'll need a little bit of help to decipher these regexps properly:
1) I think that the "24[^ ] " means "24 occurrences of any non-space character followed by a space character" -- is that right?  (if yes that looks fine to me)

2) I think that the "40[^ ] " means "40 occurrences of any non-space character" -- is that right?  (if yes that looks fine to me)

3) However what does the "%*1[ ]" means?  This is unclear to me.

Probably adding a comment here would help a lot to facilitate future maintenance.

jeannotlanglois:
>It was a preparation for returning -1 on real failures. Not having SDP is not really a failure.
>In practice, the return value is never checked, so it doesn't change any actual behavior.
>Regarding tests, aren't there unit tests for these files? I ran some manual tests, but only on 2 files.

Still that is not correct - that function is meant to only be called when SDP has been detected by the caller.  Otherwise it's a logic error.  That is why the -1 error code is for.  So you changed the behavior here.  Please fix.

orgads:
Oh, now I see that `extract_srtp_remote_info` is only called if `Content-Type` is `application/sdp`, so indeed having an empty body is probably invalid.

orgads:
Done.

PR #824, opened 2025-12-14
SRTP: Consolidate Srtp{Audio,Video}InfoParams

orgads:
They are identical.


orgads:
@jeannotlanglois Please review.

PR #828, opened 2026-01-01
Remove bundled cryptographic implementations in favor of OpenSSL/WolfSSL

lemenkov:
# Remove Bundled Cryptographic Code from SIPp

## Summary

Remove the option to build SIPp without an external SSL/TLS library. Building SIPp now requires either OpenSSL (≥1.1.1) or WolfSSL (≥3.15.0).

## Rationale

**This is a security fix, not a feature removal.**

The bundled cryptographic code:

- Originates from 1990s/2000s-era implementations
- Has never been audited
- Has no CVE tracking or security response process
- Likely contains vulnerabilities that have been fixed in upstream libraries years ago
- Lacks side-channel attack mitigations present in modern libraries

**Additional benefits:**

- Removes ~1300 lines of unmaintained code
- Eliminates duplicate code paths
- OpenSSL/WolfSSL provide hardware acceleration (AES-NI, ARMv8)
- Enables FIPS compliance for users who need it
- Thread-safety: removes global state from Rijndael implementation

**Practicality:**

SIPp is a testing tool. It runs on developer machines, CI systems, and test labs. These environments have OpenSSL. Users requiring minimal footprint can use WolfSSL, which is specifically designed for constrained systems.

## Compatibility

OpenSSL 1.1.1 was released in September 2018. Distributions still shipping older OpenSSL are either EOL or running with known security vulnerabilities. We should not compromise SIPp's security to support systems that are themselves insecure.

For reference, OpenSSL 1.1.1 is available in:

- RHEL/CentOS 8+ (2019)
- Debian 10+ (2019)
- Ubuntu 18.04+ (2018)
- Alpine 3.9+ (2019)
- Fedora 28+ (2018)


orgads (approved):
Thanks!

PR #862, opened 2026-03-29
Fix SDP extraction from multipart/mixed SIP message bodies

mt-cov:
## Summary

The Content-Type check in `process_incoming()` (call.cpp:4575) only matched `application/sdp` exactly, causing SIPp to skip SDP media extraction entirely when the body was `multipart/mixed`. This meant `rtp_stream`, PCAP play, and SRTP parameter extraction never ran for multipart messages.

## Problem

SIP messages carrying both SDP and PIDF-LO geolocation data (per RFC 5621 / RFC 5765) use `Content-Type: multipart/mixed` with the SDP in one MIME part. This is standard for E911 call flows. When SIPp received such an INVITE as a UAS, it never extracted the remote media address, so `rtp_stream` had no destination and sent zero RTP packets.

## Fix

Extend the Content-Type check to also match `multipart/*` bodies that contain an `application/sdp` part. The existing `find_in_sdp()` and `extract_rtp_remote_addr()` functions already search raw message text and work correctly with multipart content — they just were never called.

## Test plan

- SIPp UAS scenario receiving an INVITE with `Content-Type: multipart/mixed` containing SDP + PIDF-LO
- Verify `rtp_stream` sends audio (previously sent 0 packets)
- Verify plain `application/sdp` INVITEs still work as before

### DIFF
--- a/regress/github-#0862/run
+++ b/regress/github-#0862/run
@@ -1,0 +1,24 @@
+#!/bin/sh
+# This regression test is a part of SIPp.
+# Validate RTP extraction when INVITE body is multipart/mixed with SDP + PIDF-LO.
+. "`dirname "$0"`/../functions"; init
+
+uac_media_port=5072
+
+udplisten $uac_media_port >udplisten.log &
+udplisten_job=$!
+trap "kill -9 $udplisten_job 2>/dev/null" EXIT
+
+sippbg -sf uas.xml -i 127.0.0.1 -p 5070 -m 1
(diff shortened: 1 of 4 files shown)


orgads:
Thank you! Can you please add a unit test for this?

nvwx:
> Thank you! Can you please add a unit test for this?

You bet and sorry for the delay, I'll hopefully get some time to get this in this weekend.

lemenkov:
LGTM

PR #864, opened 2026-04-24
parse WAV fmt chunk, downmix multi-channel to mono, widen file sizes …

rajaamirapu:
### Problems Fixed

**Problem 1 — WAV header parser ignored the `fmt` chunk**
`get_wav_header_size()` only scanned for the start of the `data` chunk
and never read `NumChannels`, `BitsPerSample`, or `SampleRate`. Stereo
(or higher) files were forwarded as raw interleaved bytes, producing
heavily distorted audio at 2× (or worse) the correct duration.

**Problem 2 — File sizes capped at ~2 GB**
`cached_file_t::filesize`, `cached_pattern_t::filesize`, and all
playback-tracking fields in `taskentry_t`
(`audio/video_file_num_bytes`, `audio/video_file_bytes_left`,
`new_audio/video_file_size`) were declared as `int` (32-bit signed),
capping supported file size at ~2 GB — reached even sooner with
multi-channel or high-rate formats.

---

### Changes

| File | Change |
|---|---|
| `include/rtpstream.hpp` | Added `#include <stdint.h>`; widened 6 `int` fields in `taskentry_t` to `int64_t` |
| `src/rtpstream.cpp` | Replaced `get_wav_header_size()`; added `parse_wav_header()` and `wav_downmix_to_mono()`; updated `rtpstream_cache_file()` and `rtpstream_play()`; widened `filesize` in both cache structs to `int64_t` |

- **`parse_wav_header()`** — walks all RIFF chunks and extracts
  `NumChannels`, `BitsPerSample`, `SampleRate`, and the exact
  `data`-chunk byte offset from the `fmt ` chunk.
- **`wav_downmix_to_mono()`** — averages interleaved channels to mono
  at cache time; supports 8-bit unsigned and 16-bit signed PCM.
- **`cached_file_t::wav_processed` flag** — WAV stripping and downmix
  now happen once in `rtpstream_cache_file()`; `rtpstream_play()` skips
  all header logic for already-processed files.
- All processed paths produce a fresh `malloc`

### DIFF
--- a/include/rtpstream.hpp
+++ b/include/rtpstream.hpp
@@ -19,6 +19,7 @@
 #ifndef __RTPSTREAM__
 #define __RTPSTREAM__
 
+#include <stdint.h>
 #include <unordered_map>
 
 #define RTPSTREAM_MAX_FILENAMELEN 256
@@ -69,10 +70,10 @@ struct taskentry_t
     char                 *video_file_bytes_start;
     char                 *audio_current_file_bytes;
     char                 *video_current_file_bytes;
-    int                  audio_file_num_bytes;
-    int                  video_file_num_bytes;
-    int                  audio_file_bytes_left;
-    int                  video_file_bytes_left;
+    int64_t              audio_file_num_bytes;
+    int64_t              video_file_num_bytes;
+    int64_t              audio_file_bytes_left;
+    int64_t              video_file_bytes_left;
 
     /* playback timing information */
     int                  audio_ms_per_packet;
@@ -91,8 +92,8 @@ struct taskentry_t
     char                 new_video_payload_type;
     int                  new_audio_loop_count; // FILE:  <loopCount> -- PATTERN: -1 (UNUSED)
     int                  new_video_loop_count; // FILE:  <loopCount> -- PATTERN: -1 (UNUSED)
-    int                  new_audio_file_size;
-    int                  new_video_file_size;
+    int64_t              new_audio_file_size;
+    int64_t              new_video_file_size;
     char                 *new_audio_file_bytes;
     char                 *new_video_file_bytes;
     int                  new_audio_ms_per_packet;
--- a/src/rtpstream.cpp
+++ b/src/rtpstream.cpp
@@ -106,21 +106,33 @@ struct threaddata_t
 
 struct cached_file_t
 {
-    char   filename[RTPSTREAM_MAX_FILENAMELEN];
-    char   *bytes;
-    int    filesize;
+    char     filename[RTPSTREAM_MAX_FILENAMELEN];
+    char     *bytes;
+    int64_t  filesize;
+    bool     wav_processed; /* true once WAV header stripped and downmixed to mono */
 };
 
 struct cached_pattern_t
 {
-    int    id;
-    char   *bytes;
-    int  filesize;
+    int      id;
+    char     *bytes;
+    int64_t  filesize;
 };
 
 cached_file_t  *cached_files = nullptr;
 cached_pattern_t *cached_patterns = nullptr;
 int            num_cached_files = 0;
+
+/* Forward declarations for WAV utility functions defined later in this file */
+static bool  parse_wav_header(const char *data, int64_t size,
+                               int64_t *data_offset,
+                               uint16_t *num_channels,
+                               uint16_t *bits_per_sample,
+                               uint32_t *sample_rate);
+static char *wav_downmix_to_mono(const char *src, int64_t src_size,
+                                  uint16_t num_channels,
+                                  uint16_t bits_per_sample,
+                                  int64_t *out_size);
 int            next_rtp_port = 0;
 
 threaddata_t  **ready_threads = nullptr;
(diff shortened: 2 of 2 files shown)


orgads (commented):
Thanks for tackling this — the underlying problem is real, and the `uint_val()`
sign-extension fix in particular is a genuine bug fix. A few things need
addressing before this can go in, though: two memory/robustness bugs, one
silent-corruption case, and one behavior regression.

## Blockers

**1. Infinite loop on a malformed WAV**

In `parse_wav_header()`:

```c
ptr += 8 + chunk_size;
if (chunk_size & 1)
    ptr++;
```

`8 + chunk_size` is `unsigned int` arithmetic, so `chunk_size == 0xFFFFFFF8`
wraps to `0`. `chunk_size & 1` is also 0, so `ptr` never advances and the loop
spins forever on a corrupt or hostile file. The old `get_wav_header_size()` did
`ptr += 8` before adding `chunk_size`, so it always advanced by at least 8
bytes.

`chunk_size` is also never bounded against the remaining buffer, so a bogus
size walks `ptr` far past `limit` (formally UB, and on 32-bit it can wrap back
below `data` and read out of bounds). Something like:

```c
if (chunk_size > (uint64_t)(limit - ptr - 8))
    break;
ptr += 8 + (int64_t)chunk_size + (chunk_size & 1);
```

**2. Misaligned and endian-unsafe `int16_t` reads in the downmix**

`wav_downmix_to_mono()` does:

```c
const int16_t *s = reinterpret_cast<const int16_t *>(src);
```

`src` is `filecontents + data_offset`, and `data_offset` is wherever the chunk
walk happens to land — an odd-sized `LIST` chunk before `data` gives an odd
offset. This reintroduces exactly what the comment removed in this PR warned
about:

> Read as little-endian. Do not dereference as int, since it can be misaligned.

The same cast is also wrong on big-endian hosts. Please read the source samples
through the existing `uint16_val()` byte-wise accessor. The destination buffer
comes from `malloc()`, so it can keep using a typed pointer.

**3. `AudioFormat` is read about but never checked**

The comment `/* uint16 AudioFormat: PCM == 1 */` sits directly above code that
extracts channels, rate, and bit depth and then ignores the format field
entirely.

This matters most for the common SIPp case: G.711 audio streamed straight to
the wire. A stereo mu-law (format 7) or A-law (format 6) WAV reports
`bits_per_sample == 8` and lands in the 8-bit branch, where companded code
words get linearly averaged. That produces noise, not a downmix — and it
happens silently. The downmix should bail unless `AudioFormat == 1`.

## Regression

**4. Requiring both `fmt ` and `data` makes some previously working files
regress**

`parse_wav_header()` returns `false` unless both chunks are found, and the
caller then serves the raw bytes *including the RIFF header* as audio. The old
`get_wav_header_size()` stripped the header as soon as it found the `data`
chunk. So a WAVE file whose `fmt ` chunk this parser rejects (declared
`chunk_size < 16`, or a header truncated at `limit`) now plays its header as a
burst of noise where it previously played clean.

Locating `data` should be sufficient; a missing or unparseable `fmt ` can be
treated as mono with no downmix.

**5. The `data` chunk's declared size is ignored**

```c
int64_t pcm_size = (int64_t)statbuffer.st_size - data_offset;
```

Any trailing `LIST`/`INFO` chunks after `data` therefore get streamed as audio.
Clamping to `min(chunk_size, size - data_offset)` fixes it, and the parser
already has `chunk_size` in hand at that point.

## Design

**6. Peak memory doubles, which works against this PR's own goal**

The mono path `malloc()`s a full second copy of the payload and then frees the
original. For the multi-gigabyte files the `int64_t` widening is meant to
enable, that is a 2x peak. Since `cached_files[].bytes` only has to be the base
of a `malloc`'d block for `rtpstream_shutdown()`'s `free()`, a
`memmove(filecontents, pcm_start, pcm_size)` keeps the existing pointer valid
with no extra allocation at all. The downmix output is likewise always smaller
than its input, so it can run in place followed by a shrinking `realloc()`.

**7. Dead code left behind in `rtpstream_play()`**

```c
if (!cached_files[file_index].wav_processed) {
    /* Legacy path for raw (non-WAV) files: no header to strip. */
}
```

This branch has an empty body — it should just be deleted. Once it is,
`wav_processed` is only ever read inside `rtpstream_cache_file()`, so it can be
a local `bool` rather than a new field in `cached_file_t`.

**8. Failure and unsupported cases are silent**

`sample_rate` is parsed and then never used. The PR description attributes the
"2x the correct duration" symptom to channel interleaving, but a 44.1 kHz file
still plays at the wrong rate after this change, with no resampling and no
diagnostic. Similarly, when the downmix bails on 24- or 32-bit input, the code
falls through and caches the interleaved audio — the original bug, now
invisible to the user. Both cases deserve a `WARNING` log rather than silence.

## Nits

- The `(uint8_t)` casts in `uint_val()` fix a real pre-existing sign-extension
  bug in the chunk-size parse. That hunk stands on its own and carries no risk;
  it would be easy to review and land as a separate commit.
- `gtest` is already wired up in this repo (`sipp_unittest`, CMakeLists.txt).
  `parse_wav_header()` is a pure function over a byte buffer, so a few
  table-driven cases — stereo 16-bit, an odd-sized `LIST` chunk before `data`,
  and the `0xFFFFFFF8` chunk size from item 1 — would cover items 1, 4, and 5
  cheaply.
- In a C++ translation unit, prefer `<cstdint>` over `<stdint.h>` in
  `include/rtpstream.hpp`.

lemenkov:
Fully agree with Orgad's review — that covers the important ground thoroughly. One small thing to add that reinforces his point 7:

Beyond being redundant, the `wav_processed` field is actively unsafe if kept. `cached_files[]` and `cached_patterns[]` are both indexed by the same `num_cached_files` counter, but only the FILE path (`mode == 0`) ever initializes `cached_files[num_cached_files].wav_processed`. In PATTERN mode the slot is written in `cached_patterns[]` and the matching `cached_files[]` entry is left as whatever `realloc` returned — so a later read of `wav_processed` for that index is reading uninitialized memory. Collapsing it to a local `bool` as you suggest removes the field *and* this hazard in one move, so that's the way to go.

(Not for this PR, but the two parallel arrays sharing one index counter is what makes this class of bug possible — worth a cleanup someday.)

PR #867, opened 2026-05-23
feat: add startup wizard and configurable Call-ID generators

darwvin-dev:
Start an interactive startup wizard when SIPp is launched without arguments on an interactive terminal. The wizard helps build common -sn and -sf command lines, only offers options supported by the current build, and preserves the existing help behavior for non-interactive no-argument runs.

Add a new -cid_type option for built-in Call-ID generators while keeping -cid_str as the default behavior. Support UUID, compact UUID, random, and timestamp-based Call-IDs, add unit coverage for each mode, and document the new CLI behavior.

### DIFF
--- a/include/sipp.hpp
+++ b/include/sipp.hpp
@@ -165,6 +165,12 @@ cmd messages are received */
 #define DEFAULT_BEHAVIOR_PINGREPLY   4
 #define DEFAULT_BEHAVIOR_BADCSEQ     8
 
+#define CID_MODE_FORMAT              0
+#define CID_MODE_UUID                1
+#define CID_MODE_UUID_COMPACT        2
+#define CID_MODE_RANDOM              3
+#define CID_MODE_TIMESTAMP           4
+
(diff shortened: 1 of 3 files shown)


wdoekes:
Hi.
Would  you consider filing two separate PRs so they can be reviewed on their merits sepeately?
Maybe you can also take the time to document the reasoning for the added features.
Thanks!

darwvin-dev:
Hi, thanks for the feedback.
Yes, that makes sense. I’m working on splitting the changes into two separate PRs so each feature can be reviewed independently.
I’ll also add more context and document the reasoning behind each change. I’ll open the new PRs soon and link them here.
Thanks!

darwvin-dev:
Split this combined change into two focused PRs as requested:

- #868 https://github.com/SIPp/sipp/pull/868
- #869 https://github.com/SIPp/sipp/pull/869

Closing this combined PR so each feature can be reviewed on its own merits.

PR #868, opened 2026-05-23
feat: add interactive startup wizard

darwvin-dev:
Summary:
Start an interactive startup wizard when SIPp is launched without arguments on an interactive terminal. The wizard helps users build common -sn and -sf command lines and preserves the existing help behavior for non-interactive no-argument runs.

Reasoning:
Launching SIPp without arguments is a common first-touch path, especially for users who are exploring the built-in scenarios or validating a new build. Today that path drops straight to the generic help output, which is complete but still leaves the user to translate their intent into a working command line.

This wizard is meant to reduce that startup friction without changing batch or scripted behavior. It only runs when both stdin and stdout are interactive, it keeps the current help path for non-interactive usage, and it only offers scenario and transport choices that are supported by the current build.

What is included:
- Interactive prompts for common embedded and file-based scenario starts.
- Transport, remote host, request URI, call rate, and basic limit prompts.
- A final rendered command preview before execution.
- Existing no-argument help behavior retained for non-interactive runs.

What is not included:
- No Call-ID generation changes. Those are split into a separate PR for review on their own merits.


lemenkov:
I like the idea tbh. I remember myself trying it for the first time.

darwvin-dev:
Thanks for the positive feedback. I also followed up on the GitHub Advanced Security note by documenting the larger startup wizard flow in the latest commit (`469ab9f`).

Current status on the latest head:

- C/C++ CI: success
- CodeQL: success
- codespell: success

orgads:
@darwvin-dev I rebased it for you. Can you address copilot's comments?

darwvin-dev:
Thanks for the feedback. I've addressed all of the suggested changes in `a1d3142`.

I also did a local verification pass:

* `cmake --build build-pr-868-fix -j2`
* `cmake --build build-pr-868-nopcap -j2`
* `ctest --test-dir build-pr-868-nopcap --output-on-failure`

All tests passed (`59/59`)

src/sipp.cpp
```diff
@@ -264,6 +266,45 @@ static std::vector<std::string> split_simple_args(const std::string &input)
...
+        return true;
+    }
+
+    for (unsigned char c : arg) {
+        if (std::isspace(c) || c == '\'' || c == '"' || c == '\\' ||
+                c == '$' || c == '`' || c == '!' || c == '&' ||
+                c == '|' || c == ';' || c == '<' || c == '>' ||
+                c == '(' || c == ')' || c == '[' || c == ']' ||
+                c == '{' || c == '}' || c == '*' || c == '?' ||
+                c == '#') {
```
wdoekes: Where did you get this list and why is `*` not in it?

wdoekes (commented):
Quick item that caught my eye:

src/sipp.cpp
darwvin-dev: Updated this to use a conservative allow-list instead of maintaining a shell metacharacter deny-list. Simple safe characters are left unquoted, and anything outside that set is single-quoted in the preview.

Verified with:
- `cmake --build build-codex-pr-868 -j2`
- `cmake --build build-codex-pr-868-nopcap -j2`
- `ctest --test-dir build-codex-pr-868 --output-on-failure` (`61/61`)
- `ctest --test-dir build-codex-pr-868-nopcap --output-on-failure` (`59/59`)

wdoekes: Updated but not pushed?

darwvin-dev: Sorry about that. I accidentally pushed the changes to a different branch and forgot to open the PR against the main branch. It's fixed now

orgads:
Rebased, squashed to a single commit and rephrased the commit message.

orgads:
@wdoekes Any other comments or are we good to go?

orgads (approved):

PR #869, opened 2026-05-23
feat: add configurable Call-ID generators

darwvin-dev:
Summary:
Add a new -cid_type option for built-in Call-ID generators while keeping -cid_str as the default behavior. This adds UUID, compact UUID, random, and timestamp-based Call-ID modes, along with unit coverage for each mode.

Reasoning:
-cid_str is flexible and should remain the default, but it still assumes callers want to describe the Call-ID shape as a formatting pattern. In practice there are cases where a ready-made generator is more convenient or more consistent: interoperability testing against systems that expect UUID-like identifiers, quickly switching between predictable and opaque identifiers during troubleshooting, and avoiding repeated hand-built templates for common styles.

This change keeps backward compatibility by leaving -cid_str and the legacy formatting path untouched unless -cid_type is explicitly selected. The new modes provide a small set of built-in generators that cover the common cases without removing the existing customization path.

- New -cid_type CLI option.
- Built-in default, uuid, uuid-compact, random, and timestamp generator modes.
- Existing -cid_str behavior retained as the default path.
- Unit coverage for each mode.

### DIFF
--- a/include/sipp.hpp
+++ b/include/sipp.hpp
@@ -165,6 +165,12 @@ cmd messages are received */
 #define DEFAULT_BEHAVIOR_PINGREPLY   4
 #define DEFAULT_BEHAVIOR_BADCSEQ     8
 
+#define CID_MODE_FORMAT              0
+#define CID_MODE_UUID                1
+#define CID_MODE_UUID_COMPACT        2
+#define CID_MODE_RANDOM              3
+#define CID_MODE_TIMESTAMP           4
+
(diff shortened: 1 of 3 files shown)


darwvin-dev:
Follow-up on the GitHub Advanced Security note: the latest commit (`ec7ff64`) removes the custom loop-counter option parser path and reuses the existing string option parsing path, then validates `-cid_type` after option parsing.

The current GitHub Actions runs show `action_required` rather than executing jobs, so I cannot get a fresh hosted CI result until a maintainer approves the workflows for this PR.

orgads:
Since you already refactor, please C++-ify it. Make it a class, and use std::ostringstream instead of a bunch of static functions.

darwvin-dev:
Thanks for the review done in `221dbb9`: Call-ID generation is now handled by a `CallIdBuilder` class using `std::ostringstream` btw fixed the trailing `%` edge case and updated the `-cid_type` help text to match the accepted aliases.

orgads:
Thank you!

PR #873, opened 2026-06-17
Add multi-instance launcher

darwvin-dev:
## Summary
- Add a `-multi` launcher mode that reads `role,count,args` CSV rows and starts multiple SIPp child processes.
- Add `-multi_base_port` plus `{role}`, `{instance}`, `{base_port}`, `{instance_port}`, and `{port}` placeholders for generated child arguments.
- Document the CSV format and add unit coverage for parsing, validation, and argument expansion.
- Resolve CodeQL findings by canonicalizing the CSV file path, requiring a regular file, avoiding `execvp`, and using the current executable path for child launches.

## Validation
- Local compile check: `g++ -std=c++17 -Wall -Werror -pedantic -Iinclude -c src/multi_instance.cpp -o /tmp/multi_instance.o`
- Local Docker Debian build from a fresh build directory: `cmake -S /work -B /work/build-upstream -DUSE_GSL=0 -DUSE_SCTP=0 -DUSE_PCAP=0`
- Local unit tests: `/work/build-upstream/sipp_unittest` (`61 passed`)
- Local binary build: `cmake --build /work/build-upstream --target sipp`
- Local smoke: `./build-upstream/sipp -multi /tmp/sipp-multi-upstream-smoke.csv -multi_base_port 5070` with 2 UAS + 2 UAC `-m 0` children exited successfully.
- GitHub checks passing: lint, build, build-system-gtest, build-osx, build-static, build-wolfssl, codespell, CodeQL Analyze, and CodeQL alerts.

### DIFF
--- a/docs/multi_instance.rst
+++ b/docs/multi_instance.rst
@@ -1,0 +1,34 @@
+Multi-instance launcher
+=======================
+
+SIPp can launch several SIPp processes from one CSV configuration file.  This
+is useful when a test needs matching groups of UAC and UAS instances.
+
+Use ``-multi`` with a CSV file:
+
+.. code-block:: bash
+
+   ./sipp -multi multi.csv -multi_base_port 5060
+
+The CSV format is:
+
+.. code-block:: text
+
+   role,count,args
+   uas,2,"-sn uas -p {instance_port} -nostdin"
+   uac,2,"-sn uac 127.0.0.1:{instance_port} -m 100 -nostdin"
+
+Each row creates ``count`` child processes.  The ``args`` field is split like
+command-line arguments and passed to each child ``sipp`` process.
+
+The following placeholders are expanded in the ``args`` field:
+
+* ``{role}``: the role column value.
+* ``{instance}``: the zero-based instance number within that role.
+* ``{base_port}``: the value passed with ``-multi_base_port``.
+* ``{instance_port}``: ``base_port + instance``.  Use this to pair UAC and UAS
+  rows by instance number.
+* ``{port}``: a globally increasing port number for every child process.
+
+The launcher waits until all children exit and returns the first non-zero child
+exit code.  If all children exit successfully, the launcher exits with zero.
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -18,6 +18,7 @@ Welcome to SIPp reference documentation!
    3PCC_extended
    controlling
    transport
+   multi_instance
    media
    statistics
    error
--- a/include/multi_instance.hpp
+++ b/include/multi_instance.hpp
@@ -1,0 +1,46 @@
+/*
+ * Multi-instance launcher support for SIPp.
+ */
+
+#ifndef __MULTI_INSTANCE__
+#define __MULTI_INSTANCE__
+
+#include <iosfwd>
+#include <string>
+#include <vector>
+
+struct MultiInstanceSpec {
+    std::string role;
+    int count;
+    std::string args;
+};
+
+struct MultiInstanceCommand {
+    std::string role;
+    int instance;
+    int port;
+    std::string executable_path;
+    std::vector<std::string> argv;
+};
+
+std::string resolve_current_executable_path();
+
+bool parse_multi_instance_csv(const std::string &csv,
+                              const std::string &source_name,
+                              std::vector<MultiInstanceSpec> *specs,
+                              std::string *error);
+
+bool parse_multi_instance_csv_file(const std::string &path,
+                                   std::vector<MultiInstanceSpec> *specs,
+                                   std::string *error);
+
+std::vector<MultiInstanceCommand>
+build_multi_instance_commands(const std::string &program_path,
+                              const std::vector<MultiInstanceSpec> &specs,
+                              int base_port);
+
+int run_multi_instance_commands(const std::vector<MultiInstanceCommand> &commands,
+                                std::ostream &out,
+                                std::ostream &err);
+
+#endif
--- a/src/multi_instance.cpp
+++ b/src/multi_instance.cpp
@@ -1,0 +1,392 @@
+/*
+ * Multi-instance launcher support for SIPp.
+ */
+
+#include "multi_instance.hpp"
+
+#include <errno.h>
+#include <limits.h>
+#include <signal.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+#ifdef __APPLE__
+#include <mach-o/dyld.h>
+#endif
+
+#include <fstream>
+#include <iostream>
+#include <sstream>
+
+static std::string trim_copy(const std::string &value)
+{
+    size_t first = value.find_first_not_of(" \t\r\n");
+    if (first == std::string::npos) {
+        return "";
+    }
+    size_t last = value.find_last_not_of(" \t\r\n");
+    return value.substr(first, last - first + 1);
+}
+
+static bool parse_csv_line(const std::string &line,
+                           std::vector<std::string> *fields,
+                           std::string *error)
+{
+    fields->clear();
+    std::string current;
+    bool in_quotes = false;
+
+    size_t i = 0;
+    while (i < line.size()) {
+        char ch = line[i];
+        if (in_quotes) {
+            if (ch == '"') {
+                if ((i + 1 < line.size()) && line[i + 1] == '"') {
+                    current.push_back('"');
+                    i += 2;
+                    continue;
+                } else {
+                    in_quotes = false;
+                }
+            } else {
+                current.push_back(ch);
+            }
+        } else if (ch == '"') {
+            in_quotes = true;
+        } else if (ch == ',') {
+            fields->push_back(trim_copy(current));
+            current.clear();
+        } else {
+            current.push_back(ch);
+        }
+        ++i;
+    }
+
+    if (in_quotes) {
+        *error = "unterminated quoted CSV field";
+        return false;
+    }
+
+    fields->push_back(trim_copy(current));
+    return true;
+}
+
+static bool split_args(const std::string &args,
+                       std::vector<std::string> *words,
+                       std::string *error)
+{
+    words->clear();
+    std::string current;
+    char quote = 0;
+    bool escaping = false;
+
+    for (char ch : args) {
+        if (escaping) {
+            current.push_back(ch);
+            escaping = false;
+            continue;
+        }
+        if (ch == '\\') {
+            escaping = true;
+            continue;
+        }
+        if (quote) {
+            if (ch == quote) {
+                quote = 0;
+            } else {
+                current.push_back(ch);
+            }
+            continue;
+        }
+        if (ch == '\'' || ch == '"') {
+            quote = ch;
+        } else if (ch == ' ' || ch == '\t' || ch == '\r' || ch == '\n') {
+            if (!current.empty()) {
+                words->push_back(current);
+                current.clear();
+            }
+        } else {
+            current.push_back(ch);
+        }
+    }
+
+    if (escaping) {
+        current.push_back('\\');
+    }
+    if (quote) {
+        *error = "unterminated quoted argument";
+        return false;
+    }
+    if (!current.empty()) {
+        words->push_back(current);
+    }
+
+    return true;
+}
+
+static void replace_all(std::string *value,
+                        const std::string &needle,
+                        const std::string &replacement)
+{
+    size_t pos = 0;
+    while ((pos = value->find(needle, pos)) != std::string::npos) {
+        value->replace(pos, needle.size(), replacement);
+        pos += replacement.size();
+    }
+}
+
+static bool canonical_regular_file_path(const std::string &path,
+                                        std::string *canonical_path,
+                                        std::string *error)
+{
+    char resolved[PATH_MAX];
+    if (!realpath(path.c_str(), resolved)) {
+        *error = "unable to resolve multi-instance file: " + path + ": " +
+                 strerror(errno);
+        return false;
+    }
+
+    struct stat st;
+    if (stat(resolved, &st) != 0) {
+        *error = "unable to stat multi-instance file: " + std::string(resolved) +
+                 ": " + strerror(errno);
+        return false;
+    }
+    if (!S_ISREG(st.st_mode)) {
+        *error = "multi-instance file is not a regular file: " +
+                 std::string(resolved);
+        return false;
+    }
+
+    *canonical_path = resolved;
+    return true;
+}
+
+std::string resolve_current_executable_path()
+{
+#ifdef __linux__
+    char path[PATH_MAX];
+    ssize_t length = readlink("/proc/self/exe", path, sizeof(path) - 1);
+    if (length > 0) {
+        path[length] = '\0';
+        return path;
+    }
+#elif defined(__APPLE__)
+    char path[PATH_MAX];
+    uint32_t size = sizeof(path);
+    if (_NSGetExecutablePath(path, &size) == 0) {
+        char resolved[PATH_MAX];
+        if (realpath(path, resolved)) {
+            return resolved;
+        }
+        return path;
+    }
+#endif
+
+    return "sipp";
+}
+
+bool parse_multi_instance_csv(const std::string &csv,
+                              const std::string &source_name,
+                              std::vector<MultiInstanceSpec> *specs,
+                              std::string *error)
+{
+    specs->clear();
+    std::istringstream input(csv);
+    std::string line;
+    int line_number = 0;
+
+    while (std::getline(input, line)) {
+        ++line_number;
+        if (!line.empty() && line[line.size() - 1] == '\r') {
+            line.erase(line.size() - 1);
+        }
+        std::string trimmed = trim_copy(line);
+        if (trimmed.empty() || trimmed[0] == '#') {
+            continue;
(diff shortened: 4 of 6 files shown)


orgads (commented):
Built the branch and ran the cases below, so these are reproduced rather than read-only concerns.

**Scope:** this is ~440 lines of new C++ (a CSV parser, a shell-word splitter and a process supervisor) for something a shell loop already does. Worth deciding whether it belongs in sipp before reviewing the details.

### Blockers

1. **`-multi_base_port` without `-multi` aborts.** The prescan consumes it but returns false when `config_path` is empty, and `SIPP_OPTION_MULTI` has no `case` in the option switch, so it falls to `default:`:
   ```
   $ ./sipp -multi_base_port 5060
   Internal error: I don't recognize the option type for -multi_base_port
   ```

2. **`count` truncation is a silent no-op.** `long count` is checked `> 0`, then `static_cast<int>`. With `count=4294967296` the launcher starts nothing, prints nothing, and exits 0. There is also no upper bound, so `uas,100000,...` is a fork bomb from a config file.

3. **All other arguments are silently ignored.** `-multi` short-circuits `main()` before option parsing, so `./sipp -multi ok.csv -m 999 -sf /nonexistent.xml` runs happily and exits 0. Please reject any argument other than `-multi`/`-multi_base_port`.

4. **A failed re-split discards every argument instead of failing** (`src/multi_instance.cpp:340`, `if (!split_args(...)) words.clear();`). Parse-time validation ran on the *unexpanded* args, so a role containing a quote breaks the second split: role `ua's` yields `argv` of just `["./sipp"]` and the child prints help. `build_multi_instance_commands()` needs a way to report this.

### Bugs

5. **The header row must be on physical line 1** (`src/multi_instance.cpp:246`): `line_number` counts skipped comment and blank lines, so a leading `#` comment gives `x.csv:2: count must be a number`. Track the first non-comment row instead. `Role,Count,Args` is also not recognised.

6. **A `fork` failure leaves the already-started children running** (`src/multi_instance.cpp:365`): the loop breaks and then waits on them, so with `-m 0` children the launcher hangs forever. `<signal.h>` is included but never used, which suggests a `kill` loop was intended.

7. **`execv` plus the `"sipp"` fallback cannot work off Linux/macOS.** `resolve_current_executable_path()` has no branch for FreeBSD or Solaris (both supported, cf. `src/auth.cpp:25`), and `execv` does not search `PATH`, so the fallback always fails. Passing `argv[0]` through would cover it.

### Cleanups

- `SIPP_OPTION_CID_TYPE 42` (`src/sipp.cpp:128`) is added and never used — unrelated leftover.
- `trim_copy` is byte-identical to `src/sipp.cpp:187`, and `split_args` overlaps `split_simple_args` (`src/sipp.cpp:263`). Please share one copy via the new header.
- `__MULTI_INSTANCE__` is a reserved identifier; repo style is `__SIPP_X_H__`.
- `trim_copy` on quoted CSV fields strips intentional inner whitespace.
- Children share the launcher's tty, so N curses screens interleave — the docs should mention `-bg` or redirection, not only `-nostdin`.

The docs and the `{instance_port}` pairing scheme look right: `-p` defaults to a random free port, so the UAC rows in the example do not collide.

darwvin-dev:
Thanks for the detailed review. I reproduced the reported cases and pushed 9fc0817 addressing the blockers, bugs, and cleanup items.

Launcher argument validation, count/process limits, single-pass argument tokenization, CSV header/whitespace handling, fork failure cleanup, and cross-platform executable resolution have all been addressed. I also removed the unrelated option constant, consolidated the shared helpers, fixed the header guard, expanded the documentation, and added regression coverage for the reproduced cases.

The updated head is green on C/C++ CI, unit tests, macOS/static/wolfSSL builds, lint, CodeQL, and codespell.

On scope: I kept launcher mode intentionally narrow — normal SIPp CLI options are rejected in launcher mode and belong in each CSV row. If maintainers would prefer this orchestration to live outside the SIPp binary entirely, I’m happy to adjust the direction.

orgads (commented):
Thanks — I built `9fc0817` and re-ran every case. All seven items are fixed and verified:

```
$ ./sipp -multi_base_port 5060       -multi_base_port requires -multi          (99)
$ ./sipp -multi huge.csv             huge.csv:2: count is out of range         (99)
$ ./sipp -multi ok.csv -m 999        Unexpected argument in -multi mode: -m    (99)
$ ./sipp -multi cap.csv              exceeds the maximum of 256 child processes (99)
$ ./sipp -multi pr.csv -multi_base_port 65535   port allocation exceeds 65535  (99)
$ ./sipp -multi q.csv                ... -key r ua's ...   (arguments intact)
$ ./sipp -multi ok.csv               leading `#` comment and `Role,Count,Args` accepted
```

Splitting the args once and substituting placeholders per word is the right fix for item 4 — it rules out the whole re-split class of problem rather than just the reported case.

### New: nested `-multi` in the CSV is a fork bomb

Launcher-only options are not rejected inside the `args` field, and the 256 cap only applies per level:

```
role,count,args
rec,2,"-multi rec.csv"
```

```
$ timeout 6 ./sipp -multi rec.csv > rec.log ; wc -l rec.log
14265        # 2^depth, still climbing when the timeout fired
```

One check at parse time rejecting `-multi`/`-multi_base_port` in `args` covers it; the docs already describe them as launcher-only.

### Remaining, lower priority

- **SIGTERM to the launcher orphans the children.** Both children were reparented to init and kept running. `Ctrl-C` only works because the tty signals the whole process group, so `kill`, systemd, or a CI `timeout` leaks processes. A SIGINT/SIGTERM/SIGHUP handler doing what `kill_and_reap_children()` already does would close this.
- `split_simple_args()` returns `{}` when `split_command_args()` fails, so an unbalanced quote silently drops every advanced option in the **wizard**. The command preview makes it visible, but surfacing the error and re-prompting would be better. (The consolidation otherwise improves the wizard, which now handles quoted arguments correctly.)
- The `Starting` line prints `port=` even when the row does not use `{port}`, naming a port the child never binds.
- `kill_and_reap_children()` sends SIGKILL only; a SIGTERM first would let children close sockets and flush logs.
- `trim_copy()` and `split_command_args()` are generic string helpers, but `sipp.cpp` now includes `multi_instance.hpp` just to reach them — a small utility header would be a better home.
- The Linux and macOS branches of `resolve_current_executable_path()` return `""` on failure instead of falling back to `resolve_from_argv0()`, which is `#ifdef`-ed out on those platforms.

The scope question is for the maintainers, not me — I have no objection to the direction if they want it in-tree.

darwvin-dev:
@orgads Follow-up is now on 833136a (fix: close multi-instance executable trust boundary).

CodeQL #425 is now closed at the execution boundary: child execv() independently resolves the current SIPp process image from OS process metadata and no longer consumes caller-controlled argv[0] or command executable data. A regression test verifies that spoofed argv[0] values cannot influence executable resolution.

Linux/macOS/FreeBSD/Solaris use OS-backed executable resolution; unsupported or unresolved platforms fail closed.

C/C++ CI, unit/integration tests, macOS/static/wolfSSL builds, lint, codespell and CodeQL are all green on 833136a. GitHub Advanced Security has also marked CodeQL #425 resolved/outdated.

Would appreciate a final re-review when convenient.

orgads (commented):
Built and ran `833136a`. All round-3 items check out:

```
$ ./sipp -multi rec.csv     launcher-only option -multi is not allowed ...  (99)
$ ./sipp -multi q.csv       Starting ua's[0]: ...   (no bogus port= label)
$ kill -TERM <launcher>     both children gone, launcher exit=143
```

Signal handling looks right, and `find_option()` matches exactly with no `-opt=value` form, so the four launcher-only spellings cover it.

### The trust-boundary commit left dead code

Resolution now happens once in `run_multi_instance_commands()`, so the path through `build_multi_instance_commands()` is vestigial:

- `resolve_current_executable_path(const char *argv0)` starts with `(void)argv0;` and a comment about keeping the parameter "for source compatibility" — it was added three commits ago in this same PR, so there is nothing to stay compatible with.
- `MultiInstanceCommand::executable_path` is written and never read.
- `build_multi_instance_commands()`'s `program_path` parameter only fills that dead field and sets `argv[0]`, which `run_*` overwrites at exec, plus an unreachable "unable to resolve SIPp executable path" check.

Suggest letting `argv` hold only the child's arguments and prepending the trusted path at exec: that removes a parameter, a struct field and two checks, and makes the existing `command.argv.empty()` guard load-bearing for `uas,1,""`.

### Regression: Cygwin

`-multi` now fails closed on Cygwin — there is no `__CYGWIN__` branch, so it reaches `#else return "";` and exits with *unable to resolve trusted SIPp executable path*. Cygwin is supported (`src/rtpstream.cpp:316`) and provides `/proc/self/exe`, so `|| defined(__CYGWIN__)` on the Linux branch fixes it.

On CodeQL #425 itself: SIPp is not setuid, so anyone who can set `argv[0]` already controls the whole command line and can exec any binary directly — `argv[0]` is not a privilege boundary here. OS-metadata resolution is still the right primary mechanism, but failing closed traded platform coverage for a theoretical finding. The `#else` branches are precisely where CodeQL has nothing to say, so restoring the `argv[0]`/PATH fallback only there keeps both properties.

### Nits

- `sa_flags = 0` (no `SA_RESTART`) is load-bearing: it is what makes `waitpid()` return `EINTR` so the shutdown check runs. Worth a comment before someone "fixes" it and reintroduces the hang.
- The one-second SIGTERM grace period is only "waits briefly" in the docs; naming the value would help, and a child with open calls may want longer.
- `restore_signal_handlers()` in the child before `execv()` is redundant, since exec resets handled signals. Harmless.

darwvin-dev:
@orgads Thanks for the detailed round-3 review. I pushed a focused cleanup commit addressing the remaining items you called out:

* removed the vestigial `MultiInstanceCommand::executable_path`, `program_path` builder parameter, and caller-controlled `argv0` resolver parameter;
* child commands now contain only child arguments, and the trusted current executable path is prepended only at the `execv()` boundary;
* added Cygwin to the `/proc/self/exe` resolver path;
* documented why the signal handlers intentionally do not use `SA_RESTART`, so `waitpid()` can return `EINTR` during shutdown;
* documented the approximately one-second SIGTERM grace period;
* updated the unit tests for the simplified command representation and executable resolver.

Commit: `916ec3e1cf47581d590079aa8cc0acda5b265ab4`

The PR is still intentionally limited to the multi-instance orchestration use case; I don't plan to add further features here.

For the remaining scope question: the motivation for keeping this in-tree rather than as a shell loop is the coordinated behavior around deterministic role/instance port allocation, paired UAC/UAS instance orchestration, child lifecycle and signal propagation, cleanup/reaping, aggregate exit status, validation/process limits, and a portable declarative configuration.

I'd appreciate an explicit maintainer decision on whether that orchestration belongs in SIPp. If the scope is acceptable, I'm happy to keep the PR focused on review/CI fixes only.

lemenkov:
First, thank you both — @darwvin-dev for the persistence and genuinely careful work across these rounds, and @orgads for the very thorough code review over the past weeks!

The use case itself is a good one: running several coordinated SIPp instances in parallel (paired UAC/UAS, deterministic per-instance ports, aggregate exit status) is genuinely useful, and the declarative CSV format you've designed for it is clean.

Where I'd like to steer this is the *implementation approach* rather than the feature. Everything the launcher does — expanding `base + instance` ports, pairing roles, forwarding signals, reaping children, aggregating exit codes — is standard process orchestration that lives naturally *outside* the SIPp binary. Keeping it in-tree as a separate `-multi` entry point means `main()` gains a second parallel startup path, and the project takes on a permanent cross-platform maintenance burden for executable-path resolution (the `/proc/self/exe`, `_NSGetExecutablePath`, `sysctl`, Solaris, Cygwin handling) — which is also the only reason the CodeQL trust-boundary finding existed in the first place. A launcher that stays a launcher never has to re-exec itself.

So I'd propose we ship this as a helper script (Python would be my suggestion) rather than C++ in the SIPp binary. It could consume the exact same CSV format and reuse the `docs/multi_instance.rst` you've already written — just invoked as e.g. `sipp-multi.py multi.csv --base-port 5060` instead of `sipp -multi`. That keeps the feature for everyone who wants it, at a fraction of the code, with no new attack surface on the test tool itself.

One thing worth thinking through together is **Windows**. The current PR is already POSIX-only (`fork`/`execv`/`sigaction`), so it wouldn't serve Windows users as-is regardless. A Python script would actually be *more* portable here — Windows has official Python support and `subprocess` works natively there — though anyone on a shell-only setup could also just run it under WSL. If Windows is a constituency we care about for this feature, that's another point in favor of the script approach.

Would you be up for reworking it in that direction? Happy to help shape the script and keep your CSV format and docs intact.

orgads:
@lemenkov Regarding Windows, sipp doesn't support mingw anyway, and it is built on MSYS/Cygwin for Windows, so this is not a real issue.

docs/multi_instance.rst
```diff
@@ -24,58 +34,62 @@ The CSV format is:
...
+once with Python's ``shlex`` parser.  Placeholder expansion happens after
+tokenization, so placeholder values containing spaces or quote characters
+remain a single argument and are never re-parsed as shell syntax.  The helper
+always uses ``subprocess`` with an argument vector; it never invokes a shell.
 
-Each row creates ``count`` child processes.  The ``args`` field is split once
-into command-line arguments before placeholders are expanded.  This keeps
-placeholder values containing spaces or quote characters as a single
-argument instead of re-parsing them as shell syntax.
+The legacy in-binary launcher options ``-multi``, ``--multi``,
```
orgads: ???

Looks like your agent referred to the unmerged changes as if they were merged. Remove all references to intermediate work.

orgads (changes requested):

docs/multi_instance.rst
darwvin-dev: Thanks — fixed. I removed all references to the intermediate implementation and squashed the branch to a single clean commit containing only the final Python helper, tests, docs, and CI changes.

All checks are green again.

orgads (approved):

orgads (commented):
Built and exercised `1bb91ab`. Behaviour checks out:

```
14 tests pass (Python 3.14)
kill -TERM <launcher>       children gone, exit 143
child ignoring SIGTERM      SIGKILLed after 1.51s, exit 143
poll loop cost              20ms CPU over 5s with two children — negligible
```

Parse-time rejection of an empty `args` field carried across, and the redesign removes the nested-launcher recursion class structurally: there is no `-multi` entry point to recurse into and the CSV cannot select the executable. The `Starting` line printing the resolved executable also removes any doubt about which binary ran.

### Items

1. **The helper is not installed.** `CMakeLists.txt:353` has only `install(TARGETS sipp DESTINATION bin)`, so `make install` does not ship `sipp-multi.py`, and the documented default ("first looks for a sibling `sipp` binary") describes a layout nothing creates. Needs `install(PROGRAMS sipp-multi.py DESTINATION bin)`.

2. **`resolve_sipp_executable()`'s explicit-path branch reimplements `shutil.which`.** `which` already checks a path component directly instead of searching `PATH`, including the executable and not-a-directory checks:

   ```
   which('./sipp-multi.py') -> ./sipp-multi.py
   which('docs/index.rst')  -> None      # not executable
   which('ls')              -> /usr/bin/ls
   ```

   Lines 184-205 collapse to `found = shutil.which(os.path.expanduser(requested))` plus the existing raise, at the cost of slightly less specific error text.

3. **`build_commands()` re-validates what `parse_config()` already enforced** — `count <= 0` and the 256-child total, both already checked at parse time and reported with a `file:line:` prefix. Only the base-port and port-overflow checks are its own.

4. **Configuration errors exit 2 behind an argparse usage block:**

   ```
   usage: sipp-multi.py [-h] [--base-port BASE_PORT] [--sipp PATH] config
   sipp-multi.py: error: bad.csv:2: count must be a number
   ```

   A bad CSV file is not a command-line usage error. Printing to stderr and returning 1 (or 99, matching SIPp's `EXIT_OTHER`) would keep the message from being buried under usage text.

5. The Python test runs in both the `lint` and `build` CI jobs; once is enough for a pure-Python unit test.

6. The docs oversell the graceful phase on Windows: `Popen.terminate()` maps to `TerminateProcess`, so the one-second grace period does nothing there and children cannot flush their statistics. Worth marking that phase POSIX-only.

7. Three blank lines at `sipp-multi.py:21-23`.

darwvin-dev:
Thank you for the thorough and prompt review. I’ve addressed the points you raised. I’ll be available for the next couple of hours, so if you notice anything else, I’ll take care of it.

darwvin-dev:
I use SIPp quite a lot for testing my SIP server, so if you’re open to it, I’d also be happy to help prepare a new build/release after this is merged, including these new features

orgads:
@lemenkov Any comments before I merge?

sipp-multi.py
```diff
@@ -0,0 +1,342 @@
...
+
+    for process in active:
+        try:
+            process.wait()
+        except ChildProcessError:
+            pass
+
+
+def _normalize_returncode(returncode):
+    return 128 - returncode if returncode < 0 else returncode
```
lemenkov, line 247: This is a proper way since Python returns results inverted. E.g for `SEGV (11)` Python will return `-11`, but shell returns `128 + SEGV = 139` which is exactly equal to `128 - (-SEGV)`. I guess this worth commenting.

lemenkov (commented):
LGTM

lemenkov:
> @lemenkov Any comments before I merge?

LGTM, one cosmetic mostly note but I am fine with this.

